PR #10, opened 2017-05-10
Feature/sodalite flyway

ayang99:
also added configuration logging to app startup and cleaned up pom.xml formatting


andricDu:
You mind adding a small blurb to the readme about flyway and possibly a sample command?

PR #12, opened 2017-05-16
Feature/dcc 5583 sodalite client create manifest

KevinFHartmann:


sodalite-client/src/main/java/org/icgc/dcc/sodalite/client/cli/ClientMain.java
```diff
@@ -0,0 +1,25 @@
...
+
+  SodaliteConfig config;
+
+  @Autowired
+  ClientMain(SodaliteConfig config) {
+    this.config = config;
+  }
+
+  public void run(String... args) {
+    Command c = Command.parse(args);
```
andricDu: val

sodalite-client/src/main/java/org/icgc/dcc/sodalite/client/SpringApp.java
```diff
@@ -0,0 +1,22 @@
...
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+
+import lombok.val;
+
+@SpringBootApplication
+public class SpringApp {
+
+  public static void main(String... args) {
+    val app = new SpringApplication(SpringApp.class);
+    app.setBannerMode(Banner.Mode.OFF);
```
andricDu: 👍

sodalite-client/src/main/java/org/icgc/dcc/sodalite/client/cli/ClientMain.java
```diff
@@ -0,0 +1,25 @@
...
+import org.icgc.dcc.sodalite.client.command.Command;
+import org.icgc.dcc.sodalite.client.config.SodaliteConfig;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.*;
+import org.springframework.stereotype.*;
+
+@Component
+public class ClientMain implements CommandLineRunner {
+
+  SodaliteConfig config;
```
andricDu: private

sodalite-client/src/main/java/org/icgc/dcc/sodalite/client/command/Command.java
```diff
@@ -20,45 +20,40 @@
 import java.util.HashMap;
 import java.util.Map;
 
+import org.icgc.dcc.sodalite.client.config.SodaliteConfig;
+
+import lombok.Data;
 import lombok.val;
 
 //import lombok.extern.slf4j.Slf4j;
```
andricDu: delete these comments

sodalite-client/src/main/java/org/icgc/dcc/sodalite/client/command/ManifestCommand.java
```diff
@@ -0,0 +1,68 @@
...
+    val registry = new Registry(config);
+    String uploadId = getArgs()[1];
+    String fileName = getArgs()[2];
+    String result = registry.getRegistrationState(config.getStudyId(), uploadId);
+
+    val mapper = new ObjectMapper();
+    val root = mapper.readTree(result);
+    val m = new Manifest(uploadId);
+
+    for (val file : root.at("/payload/study/donor/specimen/sample/files")) {
```
andricDu: you should put this string into a constant.

sodalite-client/src/main/java/org/icgc/dcc/sodalite/client/command/ManifestCommand.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    val registry = new Registry(config);
+    String uploadId = getArgs()[1];
+    String fileName = getArgs()[2];
+    String result = registry.getRegistrationState(config.getStudyId(), uploadId);
+
+    val mapper = new ObjectMapper();
+    val root = mapper.readTree(result);
+    val m = new Manifest(uploadId);
```
andricDu: Check it:

```java
    Iterable<JsonNode> iter = () -> root.at(CONSTANT).iterator();
    val m = stream(iter.spliterator(),false)
        .map(file -> {
          val id = file.get("objectId");
          val fileId = id == null ? "<Invalid or missing object Id>" : id.asText();
          return new ManifestEntry(fileId, file.get("fileName").asText(), file.get("fileMd5").asText());
        })
        .collect(() -> new Manifest(uploadId), Manifest::add, (man1, man2) -> man1.getEntries().addAll(man2.getEntries()));
```

Now the cool thing here is you can wrap the lambda responsible for mapping in it's own method and reference it by a method reference, and you can also define your own collector class and reference that. The end result is that the important bits are now isolated in their own methods, easier to read, and can be tested without the glue code.

andricDu: So ideally it would look like
```java
Iterable<JsonNode> iter = () -> root.at(CONSTANT).iterator();
    val m = stream(iter.spliterator(),false)
        .map(this::parseFileJson)
        .collect(manifestCollector);
```

Assuming you might have better descriptive names than myself.

PR #14, opened 2017-05-17
OAuth2 Access Token with Scope Checking

andricDu:
Upload analysis object with good token:
screen shot 2017-05-17 at 4 47 24 pm

Upload analysis object a token that has bad scopes:
screen shot 2017-05-17 at 4 48 32 pm


sodalite-server/src/main/java/org/icgc/dcc/sodalite/server/security/StudyScopeStrategy.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Collections;
+import java.util.Set;
+
+import static java.lang.String.format;
+import static java.util.stream.Collectors.toList;
+
+@Slf4j
+public class StudyScopeStrategy {
+
+  private static final String SCOPE_STRATEGY = "sodalite.%s.%s";
```
andricDu: sodalite should also be configurable, we should agree on a config key for the server identifier / prefix.

ayang99: makes sense - this would be a "scope namespace": auth.server.scopeNamespace?

and then this would be SCOPE_STRATEGY = "%s.%s.%s"

PR #19, opened 2017-05-24
Controller Refactor + Swagger

andricDu:


andricDu:
@KevinFHartmann I've added you as a reviewer, give it a quick look over.

It's mostly the addition of swagger plus some controller logic reorganization. After this PR the next task will be consolidating the upload state logic to reduce the amount of coupling we currently have between the different services.

PR #25, opened 2017-06-03
Feature/dcc 5621 sodalite refactor entities logic

KevinFHartmann:
Making a PR before this mass of changes gets any bigger! :-)


sodalite-server/src/main/java/org/icgc/dcc/sodalite/server/utils/JsonUtils.java
```diff
@@ -18,42 +18,87 @@
...
-    return mapper;
+  @SneakyThrows
+  public static <T> T fromJson(String json, Class<T> toValue) {
+    return mapper.convertValue(mapper.readTree(json), toValue);
+  }
+
+  @SuppressWarnings("unchecked")
+  public static Map<String, Object> toMap(String json, String keyName) {
+    if (json == null || json.equals("")) {
+      json = "{}";
```
andricDu: I feel like this is one of those cases where we should have `@NonNull` annotations on the parameters and not bother with this. Unless there is an explicit reason the assumption that feeding a `null` into `toMap` should blow up.

sodalite-server/src/main/java/org/icgc/dcc/sodalite/server/utils/JsonUtils.java
```diff
@@ -18,42 +18,87 @@
...
+    }
+    try {
+      return mapper.convertValue(mapper.readTree(json), Map.class);
+    } catch (IllegalArgumentException | IOException e) {
```
andricDu: Exceptions get swallowed up and remain unused. I would argue for letting the mapper handle the checks for bad input and then handle the exceptions raised from that. 

Also at the very least log the exceptions caused with a `warn` or `error`

KevinFHartmann: Well, it was an attempt at error recovery... right now, any extra fields that are provided get automatically serialized into the "metadata" field as JSON data. 

However, it's possible that they might instead include an explicit "metadata" field, with, say, XML content; which would be invalid JSON; so I tried to handle that case by allowing a fallback where if the data we have isn't valid JSON, we make it valid, and serialize it so that people can get back what they put in...

Or we could just refuse to allow invalid JSON in the info field; or only allow them to input an "info" field, with
a single string that may/may not be valid JSON...

I was going to address that when I got into revising the validation code, which will have to change when we flatten the data structure.

sodalite-server/src/main/java/org/icgc/dcc/sodalite/server/utils/JsonUtils.java
KevinFHartmann: I was trying to recover from the situation where a null value in info field (I didn't update the test data SQL) could crash the JSON mapper. The right answer is not to let invalid data into the database in the first place; but the number of files I was changing had already gotten too high for one PR already. ;-)

sodalite-server/src/main/java/org/icgc/dcc/sodalite/server/utils/JsonUtils.java
andricDu: Interesting point about the XML. Keep `JsonUtils` clean and fail fast and move the exception catching and handling into the ` Metadata` class.

PR #26, opened 2017-06-07
untested changes

rtisma:


sodalite-server/src/main/java/org/icgc/dcc/sodalite/server/config/ExistenceConfig.java
```diff
@@ -0,0 +1,21 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+import static org.icgc.dcc.sodalite.server.service.ExistenceService.createExistenceService;
+
+@NoArgsConstructor
+@Configuration
+public class ExistenceConfig {
+
+  private static final String STORAGE_API = "https://storage.cancercollaboratory.org";
```
andricDu: This should be read from the application.yml

PR #37, opened 2017-06-19
Feature/dcc song 5636 identify duplicate submissions

KevinFHartmann:
My unit test now correctly detects duplicate submissions, so this feature is done! :-)


song-server/src/main/java/org/icgc/dcc/song/server/service/AnalysisService.java
```diff
@@ -50,21 +49,43 @@
...
   @SneakyThrows
+  public String update(String studyId, Analysis a) {
+    val id=a.getAnalysisId();
+    System.err.printf("WARNING: NOT UPDATING analysis id '%s': code not finished", id);
+    return id;
+  }
+
+  public String save(@NonNull String studyId, @NonNull Analysis analysis) {
+    val submitters = new ArrayList<String>();
+    analysis.getSample().forEach(s->submitters.add(s.getSampleSubmitterId()));
```
rtisma:
```
val submitters = analysis.getSample()
.stream()
.map(Sample::getSampleSubmitterId())
.collect(Collectors.toImmutableList())
```

song-server/src/main/java/org/icgc/dcc/song/server/service/AnalysisService.java
```diff
@@ -50,21 +49,43 @@
...
+    String id = findByBusinessKey(studyId, analysis.getAnalysisType(), submitters);
+
+    if (id == null) {
```
rtisma: instead of using null for verifying id was undefined, use the Optional class. This makes it clear that id is _intended_ to defined or undefined.
```
Optional<String> optionalId = findByBusinessKey(.....) //need to update findByBusinessKey to return Optional<String>
String id;
if (optionalId.isPresent()) {
    update(studyId, analysis)
}  else {
    return create(studyId, analysis)
}
return optionalId.get();
```

song-server/src/main/java/org/icgc/dcc/song/server/service/AnalysisService.java
```diff
@@ -75,15 +96,19 @@ public String create(String studyId, Analysis a) {
      repository.createVariantCall(experiment);
    } else {
      // shouldn't be possible if we validated our JSON first...
-     return "Analysis failed: Unknown Analysis Type";
+     throw new IllegalArgumentException("Invalid analysis type");
```
rtisma: 👍

song-server/src/main/java/org/icgc/dcc/song/server/service/AnalysisService.java
```diff
@@ -75,15 +96,19 @@ public String create(String studyId, Analysis a) {
...
    }
+   return id;
+  }
 
-  return id;
+  public String updateAnalysis(String studyId, Analysis a) {
+    assert false; // not coded yet
```
rtisma: create a ticket for this, then add a comment
`//TODO:  [DCC-1234]  implement update analysis`

song-server/src/main/java/org/icgc/dcc/song/server/service/AnalysisService.java
```diff
@@ -107,9 +132,13 @@ void addFile(String id, String fileId) {
 
   public Analysis read(String id) {
     val analysis = repository.read(id);
+    if (analysis == null) {
+      return null;
+    }
```
rtisma: it seems like this pattern of returning nulls is reoccurring. Create a ticket for now regarding refactoring null-return statements, and we can review later whether its worth the refactoring. This article explains my reasoning behind null logic: https://dzone.com/articles/java-optional-objects, as well as this one https://github.com/google/guava/wiki/UsingAndAvoidingNullExplained

song-server/src/main/java/org/icgc/dcc/song/server/service/UploadService.java
```diff
@@ -91,10 +91,13 @@ private void create(@NonNull String studyId, @NonNull String uploadId, @NonNull
 
     val json = s.getPayload();
     val analysis = JsonUtils.fromJson(json, Analysis.class);
-    val id = analysisService.create(studyId, analysis);
-    if (id == null) {
+    String id=null;
+    try {
+      id = analysisService.create(studyId, analysis);
+    } catch(IllegalArgumentException e) {
       return status(HttpStatus.INTERNAL_SERVER_ERROR,"Could not create id upload id '%id",uploadId);
```
rtisma: static import

song-server/src/main/java/org/icgc/dcc/song/server/service/AnalysisService.java
rtisma: also, on the toImmutableList() portion, statically import that

song-server/src/main/java/org/icgc/dcc/song/server/repository/mapper/AnalysisMapper.java
```diff
@@ -39,12 +39,13 @@ public Analysis map(int index, ResultSet r, StatementContext ctx) throws SQLExce
     val study = r.getString("study_id");
     val type = r.getString("type");
     val state = r.getString("state");
+    val info = r.getString("info");
```
KevinFHartmann: We've been putting all the table field names directly in the repository mapper classes up until now; is there any benefit from putting the field names into constants? If we do it, we should do it everywhere...

rtisma: yes, it should be in a AttributesNames.java class, and then any variable that has the string "study_id" should be changed to reference AttributeNames.STUDY_ID

rtisma:
looks great!

PR #40, opened 2017-06-22
Feature/add more server errors

rtisma:
- modified singTest with nicer output, and added error responses
- modified SampleService, DonorService, FileService, SpecimenService to throw ServerExceptions 
- added new ServerErrors and updated the way it creates errorIds, and added tests for it
- refactored hardcoded strings, and made error/output strings nicer
- added utility functions
- added factory methods to both SongError and ServerException
- added TODO comments with ticket numbers


singTest.sh
```diff
@@ -5,15 +5,72 @@ echo "SCRIPT_DIR = ${SCRIPT_DIR}"
...
+set -x 
+bad_a=`$SING_EXE save -u $non_existant_upload_id`
+set +x 
+
+
+non_existant_analysis_id="fake$a"
+header "[ERROR_TEST] Uploading manifect for non-existant analysisId $non_existant_analysis_id"
+set -x 
+${SING_EXE} manifest -a $non_existant_analysis_id -f manifest.txt
+set +x
```
KevinFHartmann: Instead of enabling & disabling set -x for every call to Sing, why not create a function (say "trace"), that turns
set -x on, runs a given command (such as sing), and turns it off again? 

Then, put a $TRACE_ENABLED variable before the path when you define ${SING_EXE}; if you set TRACE_ENABLED='trace', you turn on set -x. If you leave it as the empty string, you don't.

How does that sound?

KevinFHartmann: Instead of enabling & disabling set -x for every call to Sing, why not create a function (say "trace"), that turns
set -x on, runs a given command (such as sing), and turns it off again? 

Then, put a $TRACE_ENABLED variable before the path when you define ${SING_EXE}; if you set TRACE_ENABLED='trace', you turn on set -x. If you leave it as the empty string, you don't.

How does that sound?

song-server/src/main/java/org/icgc/dcc/song/server/service/DonorService.java
```diff
@@ -82,18 +84,18 @@ public String update(@NonNull Donor donor) {
     if (donorRepository.update(donor) == 1) {
       return "Updated";
     }
-    return "Failed";
+    return "Failed"; //TODO: [DCC-5644] need to properly handle this. Should an ServerException be thrown?
```
KevinFHartmann: Maybe.

We expect to update a record with new data; if nothing has changed, OR if the row we're trying to update doesn't exist, we might in either case get 0 rows modified (depending upon how our back-end database reports things to us).

If the row hasn't changed, we might not want to throw an error; otherwise, we would want to. If want to be really robust, when we get to the transaction handling stage, we could roll back the transaction unless a read gave us back identical data to the update that we tried to do... but that's not a priority for now.

If the database reports back that we've just modified more than one row, then we've got serious issues with our rows not being unique, and we really do have a problem that needs addressing.

PR #41, opened 2017-06-26
dcc-5646-blocking-endpoints

rtisma:
- added async and sync validation
- added async and sync option for upload endpoint (synchronous being the default)
- added validationDelay functionality inside valdiationService.validate() method, for testing purposes (should only be used in when the "test" profile is set in spring)
- added async and sync tests (sync tests should NEVER see the state "CREATED", where async test SHOULD see it)
- added getList sql command to UploadRepository to get multiple Upload objects for a list of upload ids
- added "-a" and "--async" boolean switch to UploadCommand in song-client that calls the async upload endpoint when defined, and the sync upload endpoint when not defined

### DIFF
--- a/song-core/src/main/java/org/icgc/dcc/song/core/utils/Debug.java
+++ b/song-core/src/main/java/org/icgc/dcc/song/core/utils/Debug.java
@@ -29,6 +29,14 @@ public class Debug {
         .skip(2);
   }
 
+  public static void sleepMs(long timeMs){
+    try {
+      Thread.sleep(timeMs);
+    } catch (InterruptedException e) {
+      e.printStackTrace();
+    }
+  }
+
   public static String generateHeader(String title, int lengthOfHeader, String symbol){
     int t = title.length();
     int n = Math.max(t+2, lengthOfHeader);
--- a/song-server/src/main/java/org/icgc/dcc/song/server/config/ValidationConfig.java
+++ b/song-server/src/main/java/org/icgc/dcc/song/server/config/ValidationConfig.java
@@ -29,6 +29,7 @@ import org.icgc.dcc.song.core.utils.JsonSchemaUtils;
 import org.icgc.dcc.song.server.validation.SchemaValidator;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Profile;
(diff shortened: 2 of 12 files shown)


song-core/src/main/java/org/icgc/dcc/song/core/utils/Debug.java
```diff
@@ -29,6 +29,14 @@
...
   }
 
+  public static void sleepMs(long timeMs){
+    try {
+      Thread.sleep(timeMs);
+    } catch (InterruptedException e) {
+      e.printStackTrace();
+    }
+  }
+
```
KevinFHartmann: 👍 😃 ++ for including the units! 😄

song-server/src/main/java/org/icgc/dcc/song/server/config/ValidationConfig.java
```diff
@@ -48,6 +49,12 @@ public SchemaValidator schemaValidator() {
   }
 
   @Bean
+  @Profile("test")
+  public Long validationDelayMs(){
+    return 500L;
```
KevinFHartmann: Get the delay from the Spring configuration, rather than hard coding it.

song-server/src/main/java/org/icgc/dcc/song/server/repository/UploadRepository.java
```diff
@@ -24,8 +24,14 @@
 import org.skife.jdbi.v2.sqlobject.SqlQuery;
 import org.skife.jdbi.v2.sqlobject.SqlUpdate;
 import org.skife.jdbi.v2.sqlobject.customizers.RegisterMapper;
+import org.skife.jdbi.v2.sqlobject.stringtemplate.UseStringTemplate3StatementLocator;
+import org.skife.jdbi.v2.unstable.BindIn;
```
KevinFHartmann: I notice that the BindIn class is in the "unstable" package. Have we checked to see how "unstable" it really is?

rtisma: removed this. Here is a message regarding BindIn from the jdbi creator himself
https://skife.org/jdbi/java/2011/12/21/jdbi_in_clauses.html

PR #47, opened 2017-07-20
Feature/dcc 5658 song submitter analysis

KevinFHartmann:


song-server/src/main/java/org/icgc/dcc/song/server/service/UploadService.java
```diff
@@ -62,16 +66,52 @@ public Upload read(@NonNull String uploadId) {
...
+
+      if (analysisSubmitterId.equals("")) {
+        // Our business rules say that we always want to create a new record if no analysisSubmitterId is set,
+        // even if the rest of the content is duplicated.
+        ids = Collections.emptyList();
+      } else {
+        ids = uploadRepository.findByBusinessKey(studyId, analysisSubmitterId);
+      }
+
+      if (ids.isEmpty()) {
```
andricDu: Is it possible that the `else` conditional branch from the previous `if` returns `null` for the ids? if so this would throw a `NPE`.

song-server/src/main/java/org/icgc/dcc/song/server/service/UploadService.java
```diff
@@ -62,16 +66,52 @@ public Upload read(@NonNull String uploadId) {
...
     String analysisType;
+    String uploadId;
+    val status = JsonUtils.ObjectNode();
+    status.put("status","ok");
+
     try {
-      create(studyId, uploadId, payload);
+      val analysisSubmitterId=JsonUtils.readTree(payload).at("/analysisSubmitterId").asText();
+      List<String> ids;
+
```
andricDu: get rid of this newline.

PR #49, opened 2017-07-21
Change: In the File table, each file is now directly associated with …

KevinFHartmann:
…an analysis Id.

Documentation: Updated METADATA_RULES to clarify behaviour for using analysisSubmitterId versus not. Reference to schemas now references the correct directory. Added a reference to docker directory.

### DIFF
--- a/song-server/src/test/java/org/icgc/dcc/song/server/service/FileServiceTest.java
+++ b/song-server/src/test/java/org/icgc/dcc/song/server/service/FileServiceTest.java
@@ -50,7 +50,7 @@ public class FileServiceTest {
   public void testReadFile() {
     val id = "FI1";
     val name = "ABC-TC285G7-A5-ae3458712345.bam";
-    val analysis_id="AN1";
+    val analysisId="AN1";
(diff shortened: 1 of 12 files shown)


song-server/src/test/java/org/icgc/dcc/song/server/service/FileServiceTest.java
```diff
@@ -93,19 +95,20 @@ public void testUpdateFile() {
 
     val study="ABC123";
     val id = "";
+    val analysis_id="AN1";
```
andricDu: 🐫  case please

song-server/src/test/java/org/icgc/dcc/song/server/service/FileServiceTest.java
```diff
@@ -50,14 +50,15 @@
   public void testReadFile() {
     val id = "FI1";
     val name = "ABC-TC285G7-A5-ae3458712345.bam";
+    val analysis_id="AN1";
```
andricDu: needs more 🐫

PR #64, opened 2017-08-21
uploadScope renamed to suffix, applied to GET

andricDu:
@KevinFHartmann  let me know if I missed something.

### DIFF
--- a/song-server/src/main/java/org/icgc/dcc/song/server/config/SecurityConfig.java
+++ b/song-server/src/main/java/org/icgc/dcc/song/server/config/SecurityConfig.java
@@ -34,7 +34,7 @@ import lombok.SneakyThrows;
 @EnableResourceServer
 public class SecurityConfig extends ResourceServerConfigurerAdapter {
 
-  @Value("${auth.server.uploadScope}")
+  @Value("${auth.server.suffix}")
   private String uploadScope;
 
   @Override
--- a/song-server/src/main/java/org/icgc/dcc/song/server/controller/AnalysisController.java
+++ b/song-server/src/main/java/org/icgc/dcc/song/server/controller/AnalysisController.java
@@ -94,7 +94,8 @@ public class AnalysisController {
    * @return A JSON object representing this analysis
    */
   @GetMapping(value = "/{id}")
-  public Analysis read(@PathVariable("id") String id) {
+  @PreAuthorize("@studySecurity.authorize(authentication, #studyId)")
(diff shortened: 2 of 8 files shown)


song-server/src/main/java/org/icgc/dcc/song/server/controller/AnalysisController.java
```diff
@@ -94,7 +94,8 @@
    * @return A JSON object representing this analysis
    */
   @GetMapping(value = "/{id}")
-  public Analysis read(@PathVariable("id") String id) {
+  @PreAuthorize("@studySecurity.authorize(authentication, #studyId)")
+  public Analysis read(@PathVariable("studyId") String studyId, @PathVariable("id") String id) {
     return analysisService.read(id);
   }
```
KevinFHartmann: Why is studyId being added here? Is it used anywhere

song-server/src/main/java/org/icgc/dcc/song/server/config/SecurityConfig.java
```diff
@@ -34,7 +34,7 @@
 @EnableResourceServer
 public class SecurityConfig extends ResourceServerConfigurerAdapter {
 
-  @Value("${auth.server.uploadScope}")
+  @Value("${auth.server.suffix}")
   private String uploadScope;
```
KevinFHartmann: Does suffix still represent a scope? If so, maybe we should call it "scope" instead of "suffix".

song-server/src/main/java/org/icgc/dcc/song/server/controller/AnalysisController.java
andricDu: It needs to be there for when reflection occurs and the `@PreAuthorize` annotation references `#studyId` in the spring expression. Otherwise you might get a good old NPE.

song-server/src/main/java/org/icgc/dcc/song/server/config/SecurityConfig.java
andricDu: `song.<studyId>.<suffix>` That was you can run multiple song servers with segregated data, scope schemes.

andricDu: The data probably wouldn't need to be segregated either given our unique ID generation.  You could have different SONG microservices talking to a single persistence backend but unable to touch each others data due to the `suffix` being injected at startup.

PR #67, opened 2017-08-29
Feature/dcc 5677 suppress command

rtisma:
added suppress command and add suppress test to the end of singTest


song-client/src/test/java/org/icgc/dcc/song/client/SpringAppTest.java
```diff
@@ -40,7 +33,31 @@ public void testUpload(){
...
-    executeMain("status", "-u", "UP-8b871d42-f86e-458d-8e4d-7e6688c597f9");
+    executeMain("status", "-u", "UP-826f10c8-7933-4160-9530-d24de40890ba");
+    log.info("sdf");
+  }
+
+  @Test
+  @SneakyThrows
+  @Ignore
+  public void testSave(){
+    executeMain("save", "-u", "UP-826f10c8-7933-4160-9530-d24de40890ba");
```
andricDu: Can you explain what the purpose of all these tests are? They don't seem to provide any benefit over just having a run profile for each command in your IDE.

rtisma: i use them to debug the client when something doesnt work as expected, but i could have also alternatively add a run profile for eachone. Should i remove all but one (as a placeholder) or remove the entire test?

andricDu: You can leave a single placeholder, remove the rest.

singTest.sh
```diff
@@ -49,3 +49,8 @@ set -x
 ${SING_EXE} manifest -a $a -f manifest.txt
 set +x 
 cat manifest.txt
+
+header "Suppressing Analysis"
```
andricDu: Remove this as well. We don't need this script to test everything, plus its suppressing something that wasn't even published.

PR #68, opened 2017-08-29
Feature/client unauthorized bug

rtisma:
- in Registry.java, wrapped all RestClient requests using tryClient, which reports "server not running", as well as "unauthorized.token" if a response has a 401 code.
- using the /isAlive endpoint to check if the server is running (like a health check)
- on server side, changed code to throw an exception when trying to read the status of an upload id that does not exist. This requires the ServerExceptionHandler class to be annotated with `@ControllerAdvice`, which springboot uses to intercept exceptions and transform them to ResponseEntities 
- added logging to file
- tested upload, status, save and manifest commands for the 3 cases (server not running, server running but incorrect token, and server running and correct token) and all of them executed as expected


song-client/src/main/java/org/icgc/dcc/song/client/register/Registry.java
```diff
@@ -128,7 +138,61 @@ public Status search(String studyId,
...
-    return restClient.get(accessToken, url);
+    return tryGet(accessToken, url);
+  }
+
+  /**
+   * Wraps the restClient calls, so that unauthorized.token errors are handled properly. In addition, if an error occurs, the availability of the song server is checked, to ensure its running.
+   * @param supplier Request method that returns a Status object
+   * @return the status
+   */
+  private Status tryClient(Supplier<Status> supplier){
```
KevinFHartmann: Why write specific error handling for just the REST client exceptions?

Put one try/catch block around the run method in ClientMethod, and you can catch and report on all the
various exception conditions in one place. That's simpler than passing around lambda functions, in my opinion.

song-client/src/main/java/org/icgc/dcc/song/client/register/Registry.java
```diff
@@ -128,7 +138,61 @@ public Status search(String studyId,
...
+    val status = new Status();
+    try {
+      return supplier.get();
+    } catch(ResourceAccessException e){
```
KevinFHartmann: Why only catch ResourceAccessException instead of RestClientException?

song-client/src/main/java/org/icgc/dcc/song/client/register/RestClient.java
```diff
@@ -119,17 +116,13 @@ public Status putOld(@NonNull String url, String json) {
       } else {
```
KevinFHartmann: Remove the putOld() function.

song-server/src/main/java/org/icgc/dcc/song/server/service/UploadService.java
```diff
@@ -63,7 +65,12 @@
   private final UploadRepository uploadRepository;
 
   public Upload read(@NonNull String uploadId) {
-    return uploadRepository.get(uploadId);
+    val upload = uploadRepository.get(uploadId);
```
KevinFHartmann: We should try to put client changes and server changes into separate PRs, if only for tracking purposes.

andricDu:
I do. Lets start with splitting this PR up

PR #69, opened 2017-08-29
enforcing minimum file count in the json schemas

rtisma:
- updated analysis json-schemas to have a minimum file count of 1
- updated the test. previously was only counting 2 errors, where there were actually 3 (file[0].fileName is the third error that was never reported)

### DIFF
--- a/song-server/src/main/resources/schemas/sequencingRead.json
+++ b/song-server/src/main/resources/schemas/sequencingRead.json
@@ -157,8 +157,9 @@
         }
       }
     },
-    "files": {
+    "file": {
       "type": "array",
+      "minItems": 1,
       "items": {
         "type": "object",
         "properties": {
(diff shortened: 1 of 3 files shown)


song-server/src/main/resources/schemas/sequencingRead.json
```diff
@@ -157,8 +157,9 @@
         }
       }
     },
-    "files": {
```
andricDu: why?

andricDu: This is a breaking change, and this is an array field anyways.

rtisma: it seems like a silent typo to me, since "files" is not in the required list and its also not a member in Analysis.java (https://github.com/icgc-dcc/SONG/blob/develop/song-server/src/main/java/org/icgc/dcc/song/server/model/analysis/Analysis.java#L36). 

If you compare to variantCall.json, "file" is used instead for listing the files.

rtisma: confirmed with @KevinFHartmann, this is a typo

PR #83, opened 2017-09-21
Feature/importer update

rtisma:
- added support for indexFiles (TBO, BAI and IDX)
- updated Server for XML
- added mongodb dependancy
- added DccMetadata (i.e XML data)  entity, dao (mongodb), configuration, converters, mongo Query Builder and dataFetcher
- added SimpleDccStorage implementation that takes a token, dcc-storage endpoint and an object id to download a file
- calculate md5sum and file size for all DccMetadatas by downloading them, and then save data into File entity


rtisma:
created pr for building, will notify when ready to merge

song-importer/src/main/java/org/icgc/dcc/song/importer/convert/DccMetadataConverter.java
```diff
@@ -0,0 +1,101 @@
...
+  private static final String _CLASS = "_class";
+  private static final String _ID = "_id";
+  private static final String ACCESS = "access";
+  private static final String CREATED_TIME = "createdTime";
+  private static final String FILENAME = "fileName";
+  private static final String GNOS_ID = "gnosId";
+  private static final String PROJECT_CODE = "projectCode";
+
+  public static DccMetadata convertToDccMetadata(Document document){
+    return createDccMetadata(
```
andricDu: I would have maybe went with a `builder` pattern for this one.

andricDu:
Logic is kinda gnarly at spots but that is most likely due to the inherent complexity of the task with all the crazy edge cases.

There were a few spots that I think a builder pattern may have been more appropriate but no need to change for this PR. Just something to be aware of.

PR #109, opened 2017-11-20
Feature/file id bug

rtisma:
- added ignoreAnalsysisIdCollisions to
controller
- added support for user defined analysisIds,
and removed analysisSubmitterId
- added guard for trying to create 2 analysis
with the same analysisId but different uploadId,
and tests
- updated IdService to resolve the correct
analysisID
- added analysisId to JsonSchema
- added tests for when id already exists in IdServer, and when trying to save same uploadId the second time
- updated Importer and refactored abit, to not rely on mongoDB, but instead just use the dcc-metadata api


andricDu:
@rtisma mind fixing the importer compilation issue.

PR #130, opened 2017-12-18
Feature/db index optimize

rtisma:
- added some benchmarking. basic time and load rate analysis by studyId
- added upload, status, save process flow  for benchmarking
- added indexing to schema.sql
- added full_table.sql which has a full blown view of all data. used for debuging, not loaded in schema.sql


rtisma:
ok to test

PR #131, opened 2017-12-18
fixed-duplicate-info-id-bug

rtisma:
Since there is no primary unique key restraint on any field in the info table, you can have 2 rows with the same info.id and info.id_type fields. It seems like this was never caught before, and now that the info_id_id_type_uindex has been applied, it catches it. The bug was during the update phase of an analysis. When updating an analysis, the files are deleted and then recreated. The file_ids from the info.id table were not deleted, so when an update was made, postgres saw it as a duplicate.

There was also an issue with the data.sql which was a typo. Again, only catchable now that the index for info_id_id_type_uindex is defined.


rtisma:
ok to test

PR #152, opened 2018-01-17
Feature/song 133 python sdk

rtisma:
- there are several places for refactoring, however it is functional and is good enough for merging. Developement will continue to refactor and address outstanding github issues


song-python-sdk/overture_song/client.py
```diff
@@ -0,0 +1,240 @@
...
+        self.__api = api
+
+    def create(self, study):
+        return self.__api.save_study(study) == '1'
+
+    def has(self, study_id):
+        val = self.__api.get_study(study_id)
+        return val[0] is not None
+
+    def read(self, study_id, recursive=False):
```
andricDu: recursive is never used

song-python-sdk/overture_song/entities.py
```diff
@@ -0,0 +1,348 @@
...
+from typing import Any, Type
+
+from dataclasses import dataclass
+
+from overture_song.validation import Validatable
+from overture_song.utils import Builder
+from typing import List
+from dataclasses import _isdataclass, asdict
+from overture_song.utils import check_type, check_state
+from overture_song.validation import non_null
```
andricDu: I think this guy is unused.

song-python-sdk/overture_song/entities.py
```diff
@@ -0,0 +1,348 @@
...
+    fileSize: int = -1
+    fileType: str = None
+    fileMd5sum: str = None
+    fileAccess: str = None
+
+    def validate(self):
+        raise NotImplemented("not implemented")
+
+    @classmethod
+    def create(cls,fileName, fileSize, fileType, fileMd5sum,
```
andricDu: argument names are not following python style guides. Should not be camelCase, use lowercase and underscores. 

https://www.python.org/dev/peps/pep-0008/

song-python-sdk/overture_song/entities.py
```diff
@@ -0,0 +1,348 @@
...
+        f.fileName = fileName
+        return f
+
+
+
+@dataclass(frozen=False)
+class Sample(Metadata, Validatable):
+    sampleId: str = None
+    specimenId: str = None
+    sampleSubmitterId: str = None
```
andricDu: camelCase is being used on a lot of these

song-python-sdk/overture_song/entities.py
```diff
@@ -0,0 +1,348 @@
...
+        s.aligned = aligned
+        s.analysisId = analysisId
+        s.libraryStrategy = libraryStrategy
+        s.insertSize = insertSize
+        s.pairedEnd = pairedEnd
+        s.referenceGenome = referenceGenome
+        return s
+
+
+
```
andricDu: got a bunch of blank lines here.

song-python-sdk/overture_song/entities.py
```diff
@@ -0,0 +1,348 @@
...
+    @classmethod
+    def builder(cls):
+        return Builder(Analysis)
+
+    @classmethod
+    def from_json(cls, json_string):
+        pass
+
+
+
```
andricDu: too many blank lines

song-python-sdk/overture_song/validation.py
```diff
@@ -0,0 +1,176 @@
...
+            _INTERNAL_DICT = Cls.__dict__[SPECIAL_FIELD]
+
+            def __init__(self, *args, **kwargs):
+
+                Cls.__init__(self, *args, **kwargs)
+                check_state(Validator.SPECIAL_FIELD in Cls.__dict__,
+                            "Decorator can only process dataclasses")
+                self._check_validator()
+
+            def _check_validator(self):
```
andricDu: This guy can be made a static method or just a function. It doesn't make use of the object instance through `self`. 

```python
@staticmethod
def _check_validator():
```

song-python-sdk/overture_song/model.py
```diff
@@ -0,0 +1,135 @@
...
+        fileName = get_required_field(d, 'fileName')
+        md5sum = get_required_field(d, 'fileMd5sum')
+        return ManifestEntry(fileId, fileName, md5sum)
+
+    def __str__(self):
+        return "{}\t{}\t{}".format(self.fileId, self.fileName, self.md5sum)
+
+
+class Manifest(object):
+    def __init__(self, analysis_id, manifest_entries=[]):
```
andricDu:
```python
def __init__(self, analysis_id, manifest_entries=[]):
```
In here your have `manifest_entries` being assigned a mutable type as a default argument. 

The issue with this is that when python interprets this class/method definition it actually evaluates the default arguments and creates a list that is then shared across all calls to this method.

This causes the following bug in your case:

```python
first = Manifest("AN1")
first.add_entry("foo")
print(first)
# prints "foo" 

second = Manifest("AN2")
print(second)
# prints "foo" again because its sharing the same default instance as all invocations of __init__


second.add_entry("bar")
print(first)

# prints "foo" and "bar" for first Manifest even though we only added "bar" to the second Manifest

```

The way you avoid this bug is with the following: 
```python
    def __init__(self, analysis_id, manifest_entries=None):
        if manifest_entries is None:
            manifest_entries = []
```

This way the empty list is only evaluated and assigned at new instance creation.

song-python-sdk/overture_song/entities.py
rtisma: CamelCase is required since the member variable names are used when automatically serializing the objects into json. By using camel case only for member variables of classes annotated with @dataclass, the code is simplified, more maintainable and cleaner

song-python-sdk/overture_song/model.py
rtisma: thank you! this probably explains a lot of odd behaviour i have had in the past

PR #177, opened 2018-02-03
Feature/song 165 swagger fix

rtisma:
To test, just run the server in spring.profiles.active=dev,test,default and then in your browser navigate to
http://localhost:8080/api-docs

Fixed Issue #165


song-server/src/main/java/org/icgc/dcc/song/server/config/SwaggerConfig.java
```diff
@@ -56,4 +68,15 @@ UiConfiguration uiConfig() {
         60000L);
   }
 
+  private ApiInfo apiInfo() {
+    return new ApiInfoBuilder()
```
andricDu: 👍

Issue #196, opened 2018-02-13
Database dump for DKFZ


andricDu:
- [x] Provide some sample data for DKFZ to test with (1-5 Analysis should be fine)
- [x] Match Normal QC  
- [x] Missing index files (#199)
- [x] Run Importer for DKFZ subset of data to generate database
- [x] Generate psql database dump and hand over data

PR #201, opened 2018-02-14
Feature/song 160 aws importer2

rtisma:
- added specimen filter config
- added function to cascade filters of the same type
- refactored UrlGenerators to use PortalQueries
- cleanup unused code
- added logging for building of some configurations


rtisma:
Related to #196

PR #270, opened 2018-05-07
Feature/song 258 replace jdbi with hibernate

rtisma:
- removed imported. Importer was heavily tied with JDBI. If the need for an importer occurs again, it will be brought back.
- replaced JDBI with Hibernate inplace. This means hibernate is not aware of relationships between the entities, as they are managed at the service layer.
- removed create endpoint for donor, sample, specimen controllers. Only the upload controller should be able to create donors, specimens and samples.


rtisma:
test this please

pom.xml
```diff
@@ -25,13 +25,13 @@
...
         <module>song-server</module>
         <module>song-client</module>
-        <module>song-importer</module>
     </modules>
 
     <parent>
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-starter-parent</artifactId>
-        <version>1.5.2.RELEASE</version>
+        <version>2.0.1.RELEASE</version>
```
andricDu: :+1:

PR #299, opened 2018-06-25
Feature/score 80 stricter publishing

rtisma:
related to overture-stack/score#80 
- updated ScoreService with isObjectExists and downloadObjectSpecification
- analysisService.publish now not only check for existence, but also md5 and size matching
- analysisService handles the case where the `objectMd5` is not defined in the objectSpecification, in which case there is an `ignoreUndefinedMd5` boolean switch to bypass it (i.e by setting to `true` the user acknowledges this is shady business)
- added thorough publish unit tests modelling practical scenarios with the storage and song servers falling out of sync
- updated the song-client with the ignore flag
- added tests for the storage service


song-server/src/main/java/org/icgc/dcc/song/server/model/enums/FileTypes.java
```diff
@@ -0,0 +1,36 @@
...
+ VCF("vcf"),
+ TBI("tbi"),
+ IDX("idx"),
+ XML("xml");
+
+ @Getter private final String extension;
+
+ public static FileTypes resolveFileType(@NonNull String fileType){
+  return Streams.stream(values())
+      .filter(x -> x.toString().equals(fileType))
```
andricDu: :+1:

song-server/src/main/java/org/icgc/dcc/song/server/service/AnalysisService.java
```diff
@@ -357,6 +362,26 @@ public AnalysisStates readState(@NonNull String id){
         .get();
   }
 
+  private Map<File, StorageObject> getStorageObjectsForFiles(String accessToken, List<File> files){
+    val map = ImmutableMap.<File, StorageObject>builder();
```
andricDu: you can replace this entire method with something like this:
```java
  private Map<File, StorageObject> getStorageObjectsForFiles(String accessToken, List<File> files){
    return files.stream().collect(toMap(f -> f, f -> storageService.downloadObject(accessToken, f.getObjectId())));
  }
```

song-server/src/main/java/org/icgc/dcc/song/server/service/AnalysisService.java
```diff
@@ -505,4 +530,61 @@ private static AbstractAnalysis instantiateAnalysis(@NonNull String analysisType
     return ANALYSIS_CLASS_MAP.get(analysisType).newInstance();
   }
 
+  private static void checkMismatchingFileSizes(String analysisId, Map<File, StorageObject> fileStorageObjectMap){
+    val mismatchingFileSizes = Lists.<String>newArrayList();
```
andricDu: Avoid mutations where possible. 
```java
  private static void checkMismatchingFileSizes(String analysisId, Map<File, StorageObject> fileStorageObjectMap){
    val mismatchingFileSizes = fileStorageObjectMap.entrySet().stream()
      .filter(e -> {
        val file = e.getKey();
        val storageObject = e.getValue();
        return !file.getFileSize().equals(storageObject.getFileSize());
      }).map(e -> e.getKey().getObjectId()).collect(toList());

    checkServer(mismatchingFileSizes.isEmpty(), AnalysisService.class,
        MISMATCHING_STORAGE_OBJECT_SIZES,
        "The following file objectIds have mismatching object sizes in the storage server. "
            + "The analysisId '%s' cannot be published until they all match." , analysisId);
  }
```

song-server/src/main/java/org/icgc/dcc/song/server/service/AnalysisService.java
```diff
@@ -505,4 +530,61 @@ private static AbstractAnalysis instantiateAnalysis(@NonNull String analysisType
...
+
+  private static void checkMismatchingFileMd5sums(String analysisId, Map<File, StorageObject> fileStorageObjectMap,
+      boolean ignoreUndefinedMd5){
+    val mismatchingFileMd5sums = Lists.<String>newArrayList();
+    val undefinedMd5ObjectIds = Lists.<String>newArrayList();
+    for (val entry : fileStorageObjectMap.entrySet()){
+      val file = entry.getKey();
+      val storageObject = entry.getValue();
+      val md5Mismatch = !file.getFileMd5sum().equals(storageObject.getFileMd5sum());
+      val objectMd5Undefined = !storageObject.isMd5Defined();
```
andricDu: you might as well inline this one.

PR #309, opened 2018-07-27
Fix path for example.

mgariepy:
Example file path is not correct in exampleVariantCall.json. this will fix the
example manifest generation.


andricDu:
ok to test

andricDu:
@rtisma yours to merge after review

Issue #314, opened 2018-09-10
Error creating sample using existing specimen


rtisma:

rtisma:
fixed in #315 and #320

PR #315, opened 2018-09-12
Feature/song 314 fix existing specimen2

rtisma:
Related to #314

### DIFF
--- a/song-server/src/main/java/org/icgc/dcc/song/server/service/CompositeEntityService.java
+++ b/song-server/src/main/java/org/icgc/dcc/song/server/service/CompositeEntityService.java
@@ -49,13 +49,14 @@ public class CompositeEntityService {
 
   public String save(String studyId, CompositeEntity s) {
     String id = sampleService.findByBusinessKey(studyId, s.getSampleSubmitterId());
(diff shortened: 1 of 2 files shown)


song-server/src/main/java/org/icgc/dcc/song/server/service/CompositeEntityService.java
```diff
@@ -49,13 +49,14 @@ private static Sample buildPersistentSample(CompositeEntity s){
 
   public String save(String studyId, CompositeEntity s) {
     String id = sampleService.findByBusinessKey(studyId, s.getSampleSubmitterId());
-    s.setSampleId(id);
     if (isNull(id)) {
       val sampleCreateRequest = buildPersistentSample(s);
       s.setSpecimenId(getSampleParent(studyId, s));
       sampleCreateRequest.setSpecimenId(s.getSpecimenId());
       id = sampleService.create(studyId, sampleCreateRequest);
+      s.setSampleId(id);
```
andricDu: Is this line needed? `s` is never used after this as the method will immediately return `id` when it leaves the conditional.

andricDu:
Thanks!

PR #333, opened 2018-10-01
Feature/song 311 file basename only

rtisma:
### Server Side
- updated jsonschemas for filenames not to contain any `/` type paths.
- added schema validation tests for this

### Java Client
- added `inputDirName` parameter to manifest command which verifies the files in the payload exist in that directory. It now writes the realpath (absolute) to file instead of just the filename
- provides a nice error message for files that do not exist

### Python SDK
- updated the `ManifestClient` with new source directory argument, as well as example scripts

### Song-docker
- fixed docker build issue
- updated to score-1.4.0 score-client
- refactored

### Documentation
- added new CLI interface documentation for client and python-sdk


song-python-sdk/examples/example_upload.py
```diff
@@ -77,15 +77,23 @@ def upload(config):
...
+    # NOTE:
+    # If the files exist, then the following command can used to write the manifest to a file.
+    #
+    # output_manifest_filepath = "../manifest.txt"
+    # manifest_client.write_manifest(save_response.analysisId, source_dir, output_manifest_filepath)
+
 
 def main():
-    api_config = ApiConfig("http://my-song-server-host:8080", 'ABC123', "my-access-token", debug=True)
+    api_config = ApiConfig("http://localhost:8080", 'ABC123', "ad83ebde-a55c-11e7-abc4-cec278b6b50a", debug=True)
```
rtisma: This access token is the default access token for song-docker

Issue #300, opened 2018-06-25
Request param in saveStudy endpoint not validated with payload


rtisma:
# Problem
if you make the following request

`POST /studies/XYZ123/`

with body

```json
{
  "studyId": "ABC123"
}
```

will return a 200 status code, and a study will be created with the studyId "ABC123". 

# Solution
1. Add a checkStudyEndpoint method, to validate the studyId in the endpoint and the payload match
2. The better solution, change the endpoint to be `PUT /studies/ABC123` with a body that does not include the studyId.

rtisma:
fixed in #344

PR #339, opened 2018-10-05
Verify that study id in url matches the payload #300

d8660091:


d8660091:
ok to tesk

rtisma:
ok to test

rtisma:
add to whitelist

song-server/src/main/java/bio/overture/song/server/service/StudyService.java
```diff
@@ -60,16 +61,23 @@ public boolean isStudyExist(String studyId){
   }
 
   @Transactional
-  public String saveStudy(Study study) {
-    val id = study.getStudyId();
+  public String validatedSaveStudy(Study actualStudy, String expectedStudyId) {
+    val id = actualStudy.getStudyId();
+    checkServer(id.equals(expectedStudyId), getClass(), MISMATCHING_STUDY_IDS,
+            "The studyId in the URL '%s' should match the studyId '%s' in the payload",
+            id, actualStudy.getStudyId());
```
rtisma: on second thought, move this checkServer call to the controller, and rename the method to just saveStudy. the controller should be responsible for the validation in that context. Also change the tests to use the StudyController, and just make the call there to validate the controller errors out properly.

Issue #338, opened 2018-10-04
Optimize GetAnalysisForStudy to be faster


rtisma:
test using a view. This is breaking dcc-repository

PR #342, opened 2018-10-09
Feature/song 338 get analyses optimize

rtisma:
related to #338 

- flattened all the neccessary fields in the analysis object graph into a single row in a view called FullView
- using the Stream api, filtered, modified and collected objects and constructed the expected analysis
- although a new view was added to the sql schema, this does not constitute a breaking change since the fix requires a single `create view` command on the existing database (increase in minor version only)


andricDu:
Good job!

andricDu:
@rtisma you have a failing unit test though. Looks like missing info field?

Issue #356, opened 2018-11-05
AWS null pointer for info field error


rtisma:
on the song AWS server, there was an error when doing

```bash
curl 'https://virginia.song.icgc.org/studies/BRCA-UK/analysis?analysisStates=PUBLISHED'
```

the following error occurs
```json
{                                                                                                                                                                                                                                                                    [105/9301]
  "errorId": "unknown.error",                                                                                                                                                                                                                                                  
  "httpStatusCode": 500,                                                                                                                                                                                                                                                       
  "httpStatusName": "INTERNAL_SERVER_ERROR",                                                                                                                                                                                                                                   
  "requestUrl": "http://localhost:8080/studies/BRCA-UK/analysis?analysisStates=PUBLISHED",                                                                                                                                                                                     
  "datetime": "2018-11-05T12:37:20.218-05:00[America/Toronto]",                                                                                                                                                                                                                
  "timestamp": 1541439440218,                                                                                                                                                                                                                                                  
  "message": null,                                                                                                                                                                                                                                                             
  "debugMessage": "[ROOT_CAUSE] -> java.lang.NullPointerException: null",                                                                                                                                                                                                      
  "stackTrace": [                                                                                                                                                                                                                                                              
    "java.lang.NullPointerException",                                                                                                                                                                                                                                          
    "at java.util.TreeMap.putAll(TreeMap.java:313)",                                                                                                                                                                                                                           
    "at bio.overture.song.server.model.Metadata.addInfo(Metadata.java:81)",                                                                                                                                                                                                    
    "at bio.overture.song.server.model.Metadata.setInfo(Metadata.java:56)",                                                                                                                                                                                                    
    "at bio.overture.song.server.model.Metadata.setInfo(Metadata.java:52)",                                                                                                                                                                                                    
    "at bio.overture.song.server.converter.FullViewConverter.extractDonor(FullViewConverter.java:157)",                                                                                                                                                                        
    "at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)",                                                                                                                                                                                            
    "at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:175)",                                                                                                                                                                                            
    "at java.util.ArrayList$ArrayListSpliterator.tryAdvance(ArrayList.java:1359)",                                                                                                                                                                                             
    "at java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:126)",                                                                                                                                                                                     
[...]

PR #357, opened 2018-11-05
fix: Fixed null pointer issue with info fields

rtisma:
Closes issue #356


rtisma:
info data may or maynot exist for an id. if it doesnt exist, it implies there is none just like an empty json file. So i null value should be treated as nothing.

rtisma:
fixed

Issue #370, opened 2018-11-21
fix circle-ci for song


rtisma:
Circle ci is not building 
https://circleci.com/gh/overture-stack/SONG/370

however building for travis and jenkins

Fgerthoffert:
Remove Travis for song

Issue #363, opened 2018-11-17
Add go-client building with maven


rtisma:
The root pom.xml should have a step to build the go-client for osx, linux and windows, using the GOOS and GOARCH env variables to build the right binaries. 

This automation should be added to as a build step. As an example, refer to 
https://github.com/raydac/mvn-golang

rtisma:
@d8660091 I still see some things failing
https://circleci.com/gh/overture-stack/SONG/418?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

d8660091:
> @d8660091 I still see some things failing
> https://circleci.com/gh/overture-stack/SONG/418?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

@rtisma probably because 418 still use the old .circleci config, once it uses the circleci config file in the branch, it would be fixed.

PR #372, opened 2018-11-21
Fix circleci and build go artifacts

d8660091:
#363 #370


.circleci/config.yml
```diff
@@ -50,3 +50,40 @@ jobs:
...
+                go build -v -o binaries/song-go-client-$GOOS-${GOARCH}${suffix}
+              done
+            done
+      - run:
+          name: Test
+          command: |
+            cd song-go-client
+            go test github.com/overture-stack/SONG/song-go-client/song -v
+      - store_artifacts:
+          path: ~/repo/song-go-client/binaries
```
rtisma: @d8660091 is there anyway to use the version from the pom.xml for the binary and uploading to artifactory? For example, if the version has the regex `^(\d+)\.(\d+)\.(\d+)$`, then upload the windows, mac, and linux song-go-client binaries to dcc-binaries directory in artifactory

rtisma:
deprecating use of circle ci. closing

PR #381, opened 2018-12-12
Add application level download access token

### LINKED ISSUE #380: score /download endpoint call should use application level token
Currently, when the use the publish endpoint, we validate the file data in song with the file data in score. To get the score file data, we need to call the /download endpoint. To call this endpoint, we need download scope. The Authorization token is taken from the publish request, and then passed to the /download endpoint. Becuse of this, we require upload scope for publishing AND download scope for the score server. This is not cool.

Fix is to create another token that has download scope and it is used by song (application level) to make the /download call to score. The user should not need download scope to do a song upload.

This is related to #371 

Test case is 
user-agent call publish endpoint with `collab.upload` scope. Publish should be successful

d8660091:
fix #380


song-server/src/main/java/bio/overture/song/server/config/StorageConfig.java
```diff
@@ -40,9 +40,12 @@
   @Value("${score.url}")
   private String storageUrl;
 
+  @Value("${score.downloadAccessToken}")
```
rtisma: rename to accessToken

song-server/src/main/java/bio/overture/song/server/service/StorageService.java
```diff
@@ -58,18 +58,19 @@
...
   @SneakyThrows
   public boolean isObjectExist(@NonNull String accessToken, @NonNull String objectId) {
     return doGetBoolean(accessToken, getObjectExistsUrl(objectId));
   }
 
   @SneakyThrows
-  public StorageObject downloadObject(@NonNull String accessToken, @NonNull String objectId){
-    val objectExists = isObjectExist(accessToken, objectId);
+  public StorageObject downloadObject(@NonNull String objectId){
+    val objectExists = isObjectExist(downloadAccessToken, objectId);
```
rtisma: just remembered, this `isObjectExist` method does a `GET storage.cancercollaboratory.org/upload/<objectid>`  call and receives a true or false if the object exists. This means it requires an `upload` scope to make this call. The application level score accesstoken should have both upload and download scope, to allow for the isObjectExist check to work, as well as download from the server.

song-server/src/main/resources/application.yml
```diff
@@ -23,6 +23,7 @@ server:
   port: 8080
 
 score:
+  downloadAccessToken: "download-token"
```
rtisma: add a comment that the accessToken needs to have both upload and download scope

d8660091:
@rtisma Ready for review, manually tested.

song-server/src/main/java/bio/overture/song/server/config/StorageConfig.java
```diff
@@ -40,9 +41,12 @@
   @Value("${score.url}")
   private String storageUrl;
 
+  @Value("${score.accessToken}")
+  private String scoreAccessToken;
+
   @Bean
   public StorageService storageService(){
-    return StorageService.createStorageService(new RestTemplate(),retryTemplate,storageUrl, validationService);
+    return createStorageService(new RestTemplate(), retryTemplate, storageUrl, validationService, "Bearer " + scoreAccessToken);
```
rtisma: create a static utility method that converts an access token to an authorization header value

song-server/src/main/java/bio/overture/song/server/service/StorageService.java
```diff
@@ -58,18 +58,19 @@
...
   @SneakyThrows
   public boolean isObjectExist(@NonNull String accessToken, @NonNull String objectId) {
     return doGetBoolean(accessToken, getObjectExistsUrl(objectId));
   }
 
   @SneakyThrows
-  public StorageObject downloadObject(@NonNull String accessToken, @NonNull String objectId){
-    val objectExists = isObjectExist(accessToken, objectId);
+  public StorageObject downloadObject(@NonNull String objectId){
+    val objectExists = isObjectExist(scoreAuthorizationHeader, objectId);
```
rtisma: make `isObjectExist` internally depend on the member variable `scoreAuthorizationHeader`. After some discussion, the song user's authorization token has nothing to do with songs communication with the storage server. Therefore, all communication between the song-server and the score-server should be done with an application level access token with download scope (for downloading object specifications) and upload scope (for checking object existence). The song users authorization is handled by the song-server and must be specified in the song users authorization header of their request. Song handles the responsibility of authorizing users against its endpoints, when the storage service is used "internally" by song, which should have full access, for validating the users request data

song-server/src/main/java/bio/overture/song/server/service/StorageService.java
```diff
@@ -130,8 +131,8 @@ private String getDownloadObjectUrl(String objectId){
   }
 
   public static StorageService createStorageService(RestTemplate restTemplate,
```
rtisma: delete this, and add the `@Builder` annotation on the StorageService class.

song-server/src/main/java/bio/overture/song/server/service/StorageService.java
```diff
@@ -58,18 +58,19 @@
...
 
   @SneakyThrows
-  public StorageObject downloadObject(@NonNull String accessToken, @NonNull String objectId){
-    val objectExists = isObjectExist(accessToken, objectId);
+  public StorageObject downloadObject(@NonNull String objectId){
+    val objectExists = isObjectExist(scoreAuthorizationHeader, objectId);
     checkServer(objectExists,getClass(), STORAGE_OBJECT_NOT_FOUND,
         "The object with objectId '%s' does not exist in the storage server", objectId);
-    return convertStorageDownloadResponse(objectId, getStorageDownloadResponse(accessToken, objectId));
+    return convertStorageDownloadResponse(objectId, getStorageDownloadResponse(scoreAuthorizationHeader, objectId));
```
rtisma: likewise, make `getStorageDownloadResponse` internally dependent on the member variable `scoreAuthorizationHeader`, just like for `isObjectExist`

rtisma: essentially, no non-static method should use `scoreAuthorizationHeader` as a method parameter

PR #389, opened 2019-02-08
Adds support for .tgz filetypes as part of analysis.

andricDu:


rtisma:
related to Ticket #388

PR #396, opened 2019-02-21
Update docker.rst

blabadi:
update old references to storage server to use SCORE name


rtisma:
ok to test

blabadi:
add to whitelist

Issue #402, opened 2019-03-18
Fix error messages when authorization fails due to invalid token


alekspejovic:
Currently , song returns an HTML error message. Error message should be in JSON format.

![image](https://user-images.githubusercontent.com/12387683/54545808-c31ef580-4978-11e9-8769-71a549fe40e7.png)

KevinFHartmann:
Technically, I think the HTML is coming from NGINX, but we need the proper JSON error message to come through.

rosibaj:
-  should return a 401/403 regardless of why the token is invalid

Issue #405, opened 2019-04-04
Fix HTML Error output in SONG


rosibaj:
- SONG throws html errors instead of JSON
- SONG should output JSON as the error instead

PR #410, opened 2019-04-09
Origin/402 fix error messages when authorization fails due to invalid token

KevinFHartmann:
Related to #402 
We now report {"error":"invalid_token","error_description":"401 null"} when the token is invalid, and
{"error":"invalid_token","error_description":"Ego is DOWN"}.

Once we figure out how to do nicer error handling in general, we may be able to do better than this in terms of nice error messages.


KevinFHartmann:
Now that we have a fix for issue #405, this whole PR can be scrapped; it was just a stop-gap until we figured out the full error handling that we wanted.

PR #411, opened 2019-04-10
docker build from source

rtisma:
Simplified docker build to use the source code from the current directory and just build that


docker-compose.yml
```diff
@@ -1,14 +1,11 @@
 version: '3.2'
 services:
   server:
-    container_name: "song-server-$VERSION"
-    image: "overture/song-server:$VERSION"
+    container_name: "song-server-local"
+    image: "overture/song-server:local"
     build:
       context: ./
```
blabadi: you can create a docker file for local run (without doing maven build inside it, just copy the local built artifacts) and a docker file for the release image (this PR), because I'm noticing you are forcing the compose to always be local. this could be another issue if you see that local work flow becomes slow with docker

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+COPY --from=builder /srv/song-server/target/song-server-*-exec.jar $JAR_FILE
+
+WORKDIR $SONG_HOME
+
+CMD mkdir -p  $SONG_HOME $SONG_LOGS \
+        && java -Dlog.path=$SONG_LOGS \
+        -jar $JAR_FILE \
+        --spring.config.location=classpath:/application.yml \
+        --spring.profiles.active=prod,secure,default
```
blabadi: im not sure if the profiles should be passed as an argument but just bringing that to your attention

blabadi:
nice job :+1:

PR #414, opened 2019-04-12
Feature/405 fix html error output in song

KevinFHartmann:
Related to #405


song-core/src/main/java/bio/overture/song/core/exceptions/SongError.java
```diff
@@ -54,7 +54,7 @@
 @JsonInclude(JsonInclude.Include.ALWAYS)
 @JsonPropertyOrder({ "errorId", "httpStatusCode", "httpStatusName",
     "requestUrl", "datetime", "timestamp",  "message", "debugMessage", "stackTrace" })
-@Builder
+@Builder(toBuilder = true)
```
rtisma: :+1:

song-server/src/main/java/bio/overture/song/server/exceptions/ServerExceptionHandler.java
```diff
@@ -17,118 +17,167 @@
...
-        .build();
-    log.error(error.toPrettyJson());
-    return error.getResponseEntity();
+  private static String getHttpErrorResponse(HttpServletRequest request, HttpStatusCodeException ex) {
+    val code = ex.getStatusCode();
+    return errorResponse(request, ex, code, code.name(), ex.getMessage());
   }
 
-  @ExceptionHandler(Throwable.class)
-  public ResponseEntity<String> handleThrowable(HttpServletRequest request, HttpServletResponse response, Throwable ex){
```
rtisma: Please add these back in. It is not easy to test these cases, and i believe the song-client depends on the structured output (json) when there is an unauthorized event, which will not even get to the server code that throws a `ServerException`. `SeverException` is thrown inside the service layer. I recall a 401 being returned due to an AccessDeniedException, so thats why that exception handler exists. The `Throwable` exception handler, will nicely wrap any unknown event as an UNKNOWN_ERROR (a 500).

rtisma:
great and yes that makes sense.

PR #424, opened 2019-07-09
Copied ".env" file into root folder for docker-compose.

jaserud:
Added a copy of ".env" file in the root directory to fix access denied errors from docker-compose build.


jaserud:
add to whitelist

jaserud:
ok to test

PR #432, opened 2019-07-23
Song 427/428 register and read analysis type endpoints

rtisma:
## Tickets:
#427 
#428 

## High Level Features:
- register endpoint (upload controller) 
     - validate analysisType schema using meta-schema
     - save an analysisType schema
- read endpoints for those schemas (schema controller) to retrieve schemas for specific versions
- system-level security for registration (SystemSecurity.java)

## Details:
- SchemaController implements all the read and register functionality. The associated test is SchemaServiceTest
- UploadController calls the validation service and schema service to register analysisType
- Refactored security to use StudySecurity and SystemSecurity for authorizing endpoints


song-server/src/main/java/bio/overture/song/server/controller/SchemaController.java
```diff
@@ -16,39 +16,64 @@
...
 
+import java.util.Collection;
+
+import static bio.overture.song.core.utils.JsonUtils.readTree;
+import static bio.overture.song.core.utils.JsonUtils.toPrettyJson;
+
 @RestController
-@RequestMapping(path = "/schema")
-@RequiredArgsConstructor
+@RequestMapping(path = "/schemas")
```
andricDu: what's the point of having `/schemas` when you never call it without immediately following with `/analysis`.
Maybe you don't need to have `/schemas/analysis` and you can just stick with `/schemas/`.

song-server/src/main/java/bio/overture/song/server/controller/SchemaController.java
```diff
@@ -16,39 +16,64 @@
...
+  @ApiOperation(value = "ListAnalysisTypes",
+      notes = "Retrieves a list of registered analysisType names" )
+  @GetMapping("/analysis")
+  public Collection<String> listAnalysisTypes(){
+    return analysisTypeService.listAnalysisTypeNames();
+  }
+
+  @ApiOperation(value = "GetAnalysisTypeVersion",
+      notes = "Retrieves a specific version of a schema for an analysisType" )
+  @GetMapping("/analysis/{name}/{version}")
```
andricDu: This is not proper RESTful design. 

`GET /schemas?name=<name>&version=<version>` should return a list of size one `[<schema>]`
`GET /schemas/<id>` should return a single object `<schema>`

In all instances, the schema should have a unique identifier when returned.

song-server/src/main/java/bio/overture/song/server/controller/SchemaController.java
```diff
@@ -16,39 +16,64 @@
...
+  }
 
-  @ApiOperation(value = "ListSchemasIds",
-      notes = "Retrieves a list of registered schema ids" )
-  @GetMapping("/list")
-  public ListSchemaIdsResponse listSchemaIds(){
-    return schemaService.listSchemaIds();
+  @ApiOperation(value = "GetLatestAnalysisType",
+      notes = "Retrieves the latest version of a schema for an analysisType" )
+  @GetMapping("/analysis/{name}/latest")
```
andricDu: This should also follow the same `GET /schemas?name=<name>&order=<order>&limit=1` or something like that. 
The elements of the path in a RESTful design should be nouns.

song-server/src/main/java/bio/overture/song/server/controller/UploadController.java
```diff
@@ -95,4 +97,15 @@ Upload status(@PathVariable("studyId") String studyId, @PathVariable("uploadId")
     return uploadService.save(studyId, uploadId, ignoreAnalysisIdCollisions);
   }
 
+  @ApiOperation(value = "RegisterAnalysisType", notes = "Registers an analysisType schema")
+  @PostMapping(value = "/schemas/{analysisTypeName}",
+      consumes = { APPLICATION_JSON_VALUE, APPLICATION_JSON_UTF8_VALUE })
+  @PreAuthorize("@systemSecurity.authorize(authentication)")
+  public @ResponseBody AnalysisType register(
```
andricDu: A `POST` on a schema should be encapsulated along side the other HTTP verbs. There seems to be some weird coupling between classes here as a result of this being here.

song-server/src/main/java/bio/overture/song/server/converter/FileConverter.java
```diff
@@ -19,20 +19,46 @@
...
-  FileEntity copyFile(FileEntity file);
+
+  void updateFileEntity(FileEntity ref, @MappingTarget FileEntity fileToUpdate);
+
+  // NOTE: mapstruct cannot properly generate this method because it defaults to using the builder instead of the setters. Since the info field is not apart of the builder, it does not fully copy.
+  default FileEntity copyFile(FileEntity ref){
+    val copy = new FileEntity();
+    updateFileEntity(ref, copy);
+    return copy;
+
```
andricDu: Empty newline here. Is google formatter not running in your IDE?

song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
```diff
@@ -0,0 +1,135 @@
...
+import static org.springframework.data.domain.Sort.Direction.ASC;
+import static org.springframework.data.domain.Sort.Direction.DESC;
+import static bio.overture.song.core.exceptions.ServerErrors.ANALYSIS_TYPE_NOT_FOUND;
+import static bio.overture.song.core.exceptions.ServerErrors.MALFORMED_PARAMETER;
+import static bio.overture.song.core.exceptions.ServerException.checkServer;
+import static bio.overture.song.server.model.enums.TableAttributeNames.ID;
+
+@Slf4j
+@Service
+public class AnalysisTypeService {
```
andricDu: You'll need to provide a way to get a single analysis by ID here as related to my comments about RESTful design.

Issue #436, opened 2019-08-14
Replace assertj with junit assertions


rtisma:
Due to the issues described in overture-stack/ego#376, assertj seems to be causing slow test-compile times (on the order of 10minutes). This slows developement significantly, and so is critical enhancement.

## Detailed Description
Essentially need to replace all assertj usage with equivalent org.junit.Assert.* methods
Will result in significant speed increase.

PR #437, opened 2019-08-15
Assertj fix

rtisma:
Related to #436 

# BEFORE removing assertj
`mvn clean test-compile`:   7 minutes

# AFTER removing assertj
`mvn clean test-compile`: 16.5 seconds


andricDu:
image

Issue #454, opened 2019-09-05
Update song-client (java) to support dynamic schemas


rtisma:
Add support for dynamic schema endpoints, as well as analysis update endpoint

- Add to the current presentation

rosibaj:
deploy to QA today @rtisma

PR #468, opened 2019-09-27
Song 454 update client

rtisma:
Related to #454 

# all
- cleaned up pom files. Only code that is used, is a direct dependency and not a transitive dependency. This is best practice and ensures the code doesnt break when a different direct dependency is upgraded

- formatting by maven 

# song-java-sdk
- added java-sdk module. This makes testing easier and allows for reusability when another java service needs to make requests to song.

- added `sensible` defaults to the sdk. DefaultRestClient, DefaultRestConfig...etc which are default implementations of the RestClient and RestConfig interfaces, respectively

- added RestClient tests

# song-client 
- removed spring from the client. Using a small Factory now to build it. The `Main` class is the entry point. Turns out Spring's component scan slowed down program run time by 2 seconds.

- removed useless code

- refactored client to define the CLI and uses the java-sdk to make requests

- client `*Command.java` files were refactored to use the sdk as opposed to make the calls directly. Everything else (include switches) remain the same

- Added 6 new Commands: `ListAnalysisTypesCommand`, `PingCommand`, `RegisterAnalysisTypeCommand`, `GetAnalysisTypeCommand`, `SubmitCommand`, UpdateAnalysis

- Deleted 2 commands: `SchemaCommand` `UploadCommand`, 

- ripped out `infoSearch` feature from the `SearchCommand`

- added tests to the client (finally!)

# song-core

- moved DTOs used by the java-sdk to the song-core module. These DTOs are shared between the server and sdk and client.

- added and updated some commonly used utilities

# song-server
- change the output of the ListAnalysisTypes endpoint to use a simpler PageDTO as opposed to Spring's Page


song-client/src/main/java/bio/overture/song/client/command/ListAnalysisTypesCommand.java
```diff
@@ -0,0 +1,142 @@
+/*
+ * Copyright (c) 2018. Ontario Institute for Cancer Research
+ *
```
KevinFHartmann: 2019

song-client/src/main/java/bio/overture/song/client/util/EnumConverter.java
```diff
@@ -0,0 +1,29 @@
...
+  public E convert(@NonNull String value) {
+    val enumSet = EnumSet.allOf(getEnumClass());
+    return enumSet.stream()
+        .filter(x -> x.toString().equals(value))
+        .findFirst()
+        .orElseThrow(
+            () ->
+                new ParameterException(
+                    format(
+                        "Value '%s' cannot be converter to %s. Available values are: [%s]",
```
KevinFHartmann: "converted" :

song-client/src/main/java/bio/overture/song/client/util/EnumConverter.java
```diff
@@ -0,0 +1,29 @@
...
+                        value, getEnumClass().getSimpleName(), COMMA.join(enumSet))));
+  }
```
KevinFHartmann: Why not use Enum's valueOf?

PR #506, opened 2019-11-13
Feature/170 spring profile rework

Magnitus-:
Changes:
- Created "local" profile for the local. Removed "secure" profile. Moved its configuration to the "local" profile and added a property check for the security feature flag part. Copied the configuration in the "prod" profile to the local and changed the docker-compose to only use the default and local profiles
- Added missing dependendencies in the POM files


Magnitus-:
See stable branch PR instead.

Issue #503, opened 2019-11-13
Remove dcc-common & dcc-id dependency


rtisma:
dcc-id is no longer needed, however when it is removed, several ClassNotFound errors occur at runtime. The errors are related to jaxb and it seems that song-server is dependent on transients  of the dcc-id-client dependency. dcc-common is no longer needed in java11 as the usage is minimal and can be replicated in song's utils.  This is a cleanup task.  

- Remove the depenency 
- Investigate which transient dependencies are missing after removal and include them 
- Make sure it can build after removal

PR #507, opened 2019-11-14
Song 492 federated id service

rtisma:
- update dcc-id docker component
- added dcc-id-client dependency because of a transient dependency. Out of scope for this ticket, but can be found at #503 
- added IdProperties for modelling config in application.yml
- updated IdConfig to consume IdProperties and instantiate different implementations of IdService and TokenService. This type of configuation was chosen over `@ConfigurationalOnProperty` since it is more explicit and easier to debug and understand. 
- created RestClient, which is a simplified abstraction that uses an http client and retry mechanism to execute get requests.
- added UriResolver for expanding URI templates
- added tests


song-server/src/test/resources/documents/uriResolver/good.json
```diff
@@ -0,0 +1,11 @@
...
+  "donor" : "https://example.org?sid={submitterId}&proj={studyId}",
+  "specimen" : "https://example.org?sid={submitterId}&proj={studyId}",
+  "sample" : "https://example.org?sid={submitterId}&proj={studyId}",
+  "file" : "https://example.org?anid={analysisId}&fname={fileName}",
+  "analysis":{
+    "existence" : "https://example.org?aid={analysisId}",
+    "save" : "https://example.org?aid={analysisId}",
+    "generate" : "https://example.org/generate"
+  }
+}
```
andricDu: missing newline at EOF

Issue #494, opened 2019-10-31
Remove the option for user supplied analysisIds


rtisma:
Spec: https://wiki.oicr.on.ca/pages/viewpage.action?pageId=128222479

Random Unique analysisIds will be generated locally, instead of on the federated Id server. This simplifies the interface with the federated id server. Since analysisIds are generated via UUID4, they can be generated locally on song-server side. This means the federated id server will only generate ids for:
- donor
- specimen
- sample
- file

File ids should also be generated locally, since they are UUID5 hashes of analysisId+fileName, but need to discuss this, as it assumes all fileIds are uuid

- remove analysis id from payload
- remove endpoint calling for analysis id 
- update tests for uuid.

PR #533, opened 2019-11-22
Song 494 remove analysis ids

rtisma:
- removed analysisIds
Reference #494


song-server/src/main/java/bio/overture/song/server/service/AnalysisService.java
```diff
@@ -134,21 +129,18 @@
...
 
-    // This doesnt commit the id to the id server
-    val candidateAnalysisId =
-        resolveCandidateAnalysisId(inputAnalysisId, ignoreAnalysisIdCollisions);
+    val analysisId = idService.generateAnalysisId();
 
     // Prevent duplicate analyses from being created
     checkServer(
-        !isAnalysisExist(candidateAnalysisId),
+        !isAnalysisExist(analysisId),
```
KevinFHartmann: In the (very unlikely) even that there is a random number collision, there's nothing the user can do about it.
Either remove the check, or, if you're paranoid enough, wrap it in a while(collision) { generate_new_random_number()} type loop.

song-server/src/main/java/bio/overture/song/server/controller/UploadController.java
```diff
@@ -56,20 +57,7 @@ public SubmitResponse submit(
       @RequestHeader(value = AUTHORIZATION, required = false) final String accessToken,
       @PathVariable("studyId") String studyId,
       @RequestBody @Valid String json_payload) {
-    return uploadService.submit(studyId, json_payload, false);
+    return uploadService.submit(studyId, json_payload);
```
KevinFHartmann: Should check to make sure the client code and sdk examples match the server changes.

Specifically, I noticed song-python-sdk/examples/example_upload.py, line 76  still has "ignore_analysis_id_collisions" as a parameter to api.save.

rtisma: thank you for the reminder. I completely forgot lol

PR #548, opened 2019-12-03
updated federatedIdService to locally generate objectids

rtisma:


song-server/src/main/java/bio/overture/song/server/service/FileService.java
```diff
@@ -52,13 +53,13 @@ public String create(
       @NonNull String analysisId, @NonNull String studyId, @NonNull FileEntity file) {
     studyService.checkStudyExist(studyId);
 
-    val result = idService.getFileId(analysisId, file.getFileName());
+    val result = idService.getObjectId(analysisId, file.getFileName());
```
KevinFHartmann: What kinds of objects can an objectId identify?

If it's only "files and index files", we should keep the name as FileId.

If it's "potentially anything", then we should drop the word "object" -- it's just a generic id.

If it's something else, we should try to identify what that something else is, and name it that?

Just a thought...

rtisma: yea it should be FileId, but was trying to conform to the icgc notion of objectId. ill change it back to FileId, and possibly in the future change the objectId field in the File entity to fileId.

Issue #481, opened 2019-10-09
BUG - registerAnalysis endpoint does not block analysisState field


rtisma:
currently, if you create a registerAnalysis request with `"analysisState" : "PUBLISHED"` you can create an analysis with a PUBLISHED state...which should not be allowed since this warrants corruption.

You should only be able to register the non-system data

- [ ] Add blocking in the json schema to analysisType.metaschema.json
- [ ] Create tests with fixtures, much like this [test](https://github.com/overture-stack/SONG/blob/develop/song-server/src/test/java/bio/overture/song/server/controller/AnalysisControllerTest.java#L314-L332)

PR #549, opened 2019-12-03
song-481-register-analysis-state-bug

rtisma:
added non-analysisState field constraint to registration metaschema and test

Since its a bug fix, increase patch number, so new version is 3.0.2-SNAPSHOT

rellated to #481

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,7 +21,7 @@
     <groupId>bio.overture</groupId>
     <artifactId>song</artifactId>
     <packaging>pom</packaging>
-    <version>3.0.2-SNAPSHOT</version>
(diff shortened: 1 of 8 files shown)


pom.xml
```diff
@@ -21,7 +21,7 @@
     <groupId>bio.overture</groupId>
     <artifactId>song</artifactId>
     <packaging>pom</packaging>
-    <version>3.0.1-SNAPSHOT</version>
+    <version>3.0.2-SNAPSHOT</version>
```
KevinFHartmann: I thought we said we were only going to increment version numbers for a release to the master branch?

rtisma: yes you are right, i completely forgot!

PR #553, opened 2019-12-04
Enhancement: CreateStudy endpoint now returns a JSON object to keep S…

KevinFHartmann:
…wagger happy.


song-server/src/main/java/bio/overture/song/server/service/StudyService.java
```diff
@@ -72,7 +73,7 @@ public String saveStudy(Study study) {
         study);
     studyRepository.save(study);
     infoService.create(id, study.getInfoAsString());
-    return id;
+    return format("{ \"message\": \"Successfully created study '%s'\" }", id);
```
rtisma: I would move this to the StudyController since this is a specific presentation of the data. Then at the controller, replace `public String saveStudy...` with `public GenericMessage saveStudy...` where generic message looks like:
```java
@Value
public class GenericMessage {
    @NonNull private final String message;
}
```

PR #558, opened 2019-12-09
Bugfix: sing now computes it's parent directory correctly

KevinFHartmann:
Enhancement: docker client now runs as a song user that has the sing binary in the PATH environment variable

### DIFF
--- a/docker/tools/score-client-demo
+++ b/docker/tools/score-client-demo
@@ -6,4 +6,4 @@ BASH_SCRIPT_DIR=$( dirname  "${BASH_SCRIPT}")
 DOCKERFILE_NAME=Dockerfile \
 	docker-compose \
 	-f ${BASH_SCRIPT_DIR}/../../docker-compose.yml \
-	run --rm score-client score-client $@
+	run --rm score-client bin/score-client $@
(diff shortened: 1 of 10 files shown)


docker/tools/score-client-demo
```diff
@@ -6,4 +6,4 @@ BASH_SCRIPT_DIR=$( dirname  "${BASH_SCRIPT}")
 DOCKERFILE_NAME=Dockerfile \
 	docker-compose \
 	-f ${BASH_SCRIPT_DIR}/../../docker-compose.yml \
-	run --rm score-client bin/score-client $@
+	run --rm score-client score-client $@
```
rtisma: bit of a chicken-and-egg problem here, which is why this command does not work. The docker container for score-client doesnt have the executable on the class path yet. that fix is in this pr: https://github.com/overture-stack/score/pull/215

for consistency sake, change this line back to
```bash
run --rm score-client bin/score-client $@
```

and in a new PR, we can update the version of the score-client/score-server docker images and then use the executable name only

docker/tools/score-client-dev
```diff
@@ -6,4 +6,4 @@ BASH_SCRIPT_DIR=$( dirname  "${BASH_SCRIPT}")
 DOCKERFILE_NAME=Dockerfile.dev \
 	docker-compose \
 	-f ${BASH_SCRIPT_DIR}/../../docker-compose.yml \
-	run --rm score-client bin/score-client $@
+	run --rm score-client score-client $@
```
rtisma: refer to above comment

PR #574, opened 2020-01-07
Remove dcc common and dcc id dependency#503

KevinFHartmann:
Now compiles, builds, and passes make test-*.


song-server/pom.xml
```diff
@@ -240,7 +229,11 @@
             <groupId>javax.xml.bind</groupId>
             <artifactId>jaxb-api</artifactId>
         </dependency>
-
+       <dependency>
+           <groupId>org.glassfish.jaxb</groupId>
+           <artifactId>jaxb-runtime</artifactId>
+           <version>2.3.1</version>
```
rtisma: move this version definition into the `<dependencyManagement>` block of the root pom.xml. If this dependency is only used in song-server, then rip out the dependency from the dependencyManagement section and put it here. Also, use the version variable like `${jaxb.version}`

song-server/src/main/java/bio/overture/song/server/controller/AnalysisController.java
```diff
@@ -82,7 +81,7 @@
       @ApiParam(value = "Non-empty comma separated list of analysis states to filter by")
           @RequestParam(value = "analysisStates", defaultValue = "PUBLISHED", required = false)
           String analysisStates) {
-    return analysisService.getAnalysis(studyId, ImmutableSet.copyOf(COMMA.split(analysisStates)));
+    return analysisService.getAnalysis(studyId, ImmutableSet.copyOf(",".split(analysisStates)));
```
rtisma: Create a `Splitters` class with a no args constructor, that contains:
`public static final Splitter COMMA = Splitter.on(",");`
and likewise for any other splits.

song-server/src/main/java/bio/overture/song/server/controller/analysisType/AnalysisTypePageableResolver.java
```diff
@@ -59,7 +58,7 @@
   public static final String DEFAULT_SORT_VARIABLE = VERSION;
   public static final Direction DEFAULT_SORT_ORDER = DESC;
 
-  private static final String ALLOWED_SORT_VARIABLES = COMMA.join(VERSION, NAME);
+  private static final String ALLOWED_SORT_VARIABLES = VERSION + "," + NAME;
```
rtisma: Create a no args constuctor `Joiners` class that contains  the following:
`public static final Joiner COMMA = Joiner.on(",");`
likewise for any other characters to join on

song-server/src/main/java/bio/overture/song/server/exceptions/ServerExceptionHandler.java
```diff
@@ -172,7 +171,7 @@ private static String errorResponseBody(
   }
 
   private static List<String> getFullStackTraceList(Throwable t) {
-    return NEWLINE.splitToList(getStackTraceAsString(t)).stream()
+    return stream(getStackTraceAsString(t).split("\n"))
```
rtisma: newline splitter

song-server/src/main/java/bio/overture/song/server/service/AnalysisService.java
```diff
@@ -415,7 +415,7 @@ private void checkMissingFiles(String analysisId, List<FileEntity> files) {
         "The following storage objectIds must be uploaded to the storage server before the "
             + "analysisId %s can be published: %s",
         analysisId,
-        COMMA.join(missingFileIds));
+        join(",", missingFileIds));
```
rtisma: use Joiners.COMMA

song-server/pom.xml
KevinFHartmann: String constants are a valid part of the Java language. Don't be so afraid of them!

If we expect the definition of a constant to ever change, then YES, by all means, we should use a name rather than an inline value.

OBJECT_NAME_DELIMITER might change. NEWLINE might change (based upon different operating systems idea of what a new line should be). 

 But the definition of a something like COMMA should never change, and if it does, it's confusing.

All we accomplish by separating the definition of the constant from the code in this case is forcing the developer to do extra work to double check that COMMA is defined as ",".

rtisma: i find that string constants are harder to read and make it hard to search for via the IDE. by using COMMA, you can find all usages which is very handy when inspecting its usage. Regarding double checking, for something that sensitive, there should be tests in place that ensure it functions as expected.

KevinFHartmann: I'm confused. How is searching for "," any harder than searching for COMMA? If anything, it should be easier, since it's two less characters to type. :-) 

There should always be tests. That doesn't mean that the code does what you think it does when you read it.:-) 

But for consistency, I'll make you happy. :-)

rtisma: thank you!

song-core/src/main/java/bio/overture/song/core/utils/Separators.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+package bio.overture.song.core.utils;
+
+import com.google.common.base.Joiner;
+import com.google.common.base.Splitter;
+import java.util.Iterator;
+import java.util.List;
+
+/** Joiners and Splitters for commonly used separators. */
+public enum Separators {
```
rtisma: love it!

PR #577, opened 2020-01-13
Fix matched normal

rtisma:
- added analysisTypeId validation via jsonSchema instead of in java code
- renamed `/schemas/meta` endpoint to `/schemas/registration` and associated method calls and variables
- made `registration` a prohibited analysisType name and added test
- added json schema constraint that `matchedNormalSubmitterSampleId` must be null when `tumourNormalDesignation != 'Tumour'`
- adjusted all fixtures for new base schema constraint


song-server/src/test/java/bio/overture/song/server/utils/EndpointTester.java
```diff
@@ -160,9 +161,9 @@ public ResponseOption getAnalysisTypeVersionGetRequestAnd(
         .getAnd();
   }
 
-  // GET /schemas/meta
-  public ResponseOption getMetaSchemaGetRequestAnd() {
-    return initWebRequest().endpoint(SLASH.join(SCHEMAS, META)).getAnd();
+  // GET /schemas/registration
```
andricDu: don't need to change now, but proper java doc on public methods if you are going to comment

PR #584, opened 2020-02-04
updated non root

henro001:
Updated the dockerfile to use non root user


rtisma:
hey the docker-compose should also be updated to run the song server as the song user

henro001:
I added the user into docker compose

rtisma:
closing infavour of #587

Issue #650, opened 2020-07-03
BUG - Transactions are not behaving as expected


rtisma:
# Problem 
During some brief investigation via intellij debugger and enabling TRACE hibernate logging, it seems that transactions are not joining higher level transactions. In addition, a test was not found that can prove/disprove transactions are working as expected.  

# Impact
This can corrupt the database. Imagine a transaction where there are several inserts, say the donorId being one of them. Mid way another insert fails for some reason. If the whole transaction is rolled back, which would incluide the donorId, then on a retry, the insert for the same donorId should be successful, since it does not exist (due to the rollback). In the current state, this rollback does not happen, which means on the retry, the insert for the same donor will FAIL. This means the database is **corrupted**.

# Work
- creating tests
- researching and experimenting. From previous experience, Hibernate is sometimes a bit of a black box, and so while researching how to do something, alot of experimentation will be needed
- most likely significant refactor. Since `@Transactional` applies to methods, some intermediate methods may need to be created. In addition, to make transactions testible, some refactor may be needed

In general, it is hard to bound the amount of work or time needed. This is more of a research project if anything since we are dealing with hibernate.

PR #651, opened 2020-07-03
song-645-events-before-tx

rtisma:
This is just a small refactor. Tests will be added as part of a discovered bug: https://github.com/overture-stack/SONG/issues/650


rtisma:
Scrapping this code, refer to https://github.com/overture-stack/SONG/pull/652

Issue #645, opened 2020-06-27
BUG - Events are sent before transactions are commited


rtisma:
## Problem
Currently, publish events are sent at the end of the publish call, which is marked with @Transactional. It has been observed (at least twice) in the genomics-data-playground, that when 4 payloads are published back to back, 1 of them is not indexed. After reviewing maestro, an error was discovered indicating the analysisId did not exist. This smells like a race condition since the event was created by song and processed by maestro BEFORE the actual state of the analysis was committed to the database as published by Song. Since the network latency between kafka and song and maestro and elastic search is the very small via genomics-data-playground (since all running on the same host), this issue was able to be replicated.

## Impact
If not fixed, there maybe missing data in the portal. The only way to ensure everything is "synched" is to manually trigger indexing for every study...which defeats the purpose of the event driven approach. This is not a breaking change.

## Solution:
rip out any call to [checkedUpdateState](https://github.com/overture-stack/SONG/blob/develop/song-server/src/main/java/bio/overture/song/server/service/AnalysisService.java#L439) inside of a transactional method, and place it after the transactional method. This will ensure the transaction completed before sending the message.

rtisma:
Additional problems:
The changes needed to refactor some of the code, however this cause hibernate persistent context issues. It seems the hibernate/spring magic is no longer working, and needs to be fixed. Even though the trasnaction annotation is still there but on a different method, the persistent context is no longer present during analysis creation, which is not exactly intuitive. The complexity of this problem has escalated and will take more time now than initially estimated

rosibaj:
@tisma deploy confirmed in standup so will close now!

PR #652, opened 2020-07-07
Song 645 events before tx, the second

rtisma:
Related to #645 

Ripped out the sending logic, and using the Decorator pattern for adding "sending" functionality to the analysisService. This was neccessary since transactions can only be triggered by EXTERNAL calls, not calls within the same object.

This is not trivial to test, since the analysisService is using injected dependencies and does not have a correct constructor. Simply verified by enabling transaction logging and ensuring transactions at committed BEFORE the sender call.

In addition, a separate ticket was created to address the lack of transaction tests https://github.com/overture-stack/SONG/issues/653


andricDu:
just a comment, but for organization, you could put all the analysis related services into an `analysis` package inside of the `service` package.

PR #661, opened 2020-07-22
song-655-jwt-2

rtisma:


song-core/src/main/java/bio/overture/song/core/exceptions/ServerErrors.java
```diff
@@ -72,6 +73,7 @@
   SEQUENCING_READ_NOT_FOUND(NOT_FOUND),
   VARIANT_CALL_NOT_FOUND(NOT_FOUND),
   UNAUTHORIZED_TOKEN(UNAUTHORIZED),
+  FORBIDDEN_TOKEN(FORBIDDEN),
```
andricDu, line 76: :+1:

song-server/src/main/java/bio/overture/song/server/oauth/ExpiringOauth2Authentication.java
```diff
@@ -7,15 +7,15 @@
 
 @Getter
 public class ExpiringOauth2Authentication extends OAuth2Authentication {
-  private final int expiry; // expiry time of the authentication token in seconds
+  private final long expiry; // expiry time of the authentication token in seconds
```
andricDu, line 10: :+1:

song-server/src/main/resources/application.yml
```diff
@@ -152,6 +152,20 @@ id.persistInMemory: true
 # This url should not exist
 score.url: http://localhost:0
 
+auth:
+  server:
+    url: "http://localhost:8084/check_token/"
+    clientId: "3kJhz9pNtC0pFHAxr2SPkUkGjXrkWWqGcnPC0vBP"
```
andricDu, line 158: I'm assuming these are not real.

song-server/src/main/java/bio/overture/song/server/oauth/ExpiringOauth2Authentication.java
andricDu: Just a note, the `AccessTokenConverterWithExpiry` is still using int for the `expiryInSeconds`

song-server/src/main/java/bio/overture/song/server/security/JWTTokenConverter.java
```diff
@@ -0,0 +1,89 @@
...
+    // TODO: rtisma --- this is also a hack. the resourceIds maps to the "aud" field. This should be
+    // empty inorder for the OAuth2AuthenticationManager to process properly
+    authentication.getOAuth2Request().getResourceIds().clear();
+
+    return authentication;
+  }
+
+  private static long calcSecondsUntilExpiry(Long expirationTimestamp) {
+    val diff = expirationTimestamp - MILLISECONDS.toSeconds(currentTimeMillis());
+    return diff > 0 ? diff : 0;
```
andricDu, line 55: Generally this is fine, and you are fine in this situation, but you can get into trouble when boxing and widening happen in the wrong order. I would personally use `0L` whenever I can because when using `val` you are boxing primitives. Again, you are fine in these situations, just something to think about.

PR #684, opened 2020-10-20
Analysis State Change History

joneubank:
- Created At and Updated At properties for analysis
- New Entity AnalysisStateChange to record analysis state history
- Calculated properties to show publish date and first publish date


song-server/src/main/java/bio/overture/song/server/model/analysis/Analysis.java
```diff
@@ -69,9 +61,20 @@
...
+  @UpdateTimestamp
+  private LocalDateTime updatedAt;
+
+  @Transient private LocalDateTime firstPublishedAt;
+  @Transient private LocalDateTime publishedAt;
+
   @NotNull
   @JsonIgnore
-  @ManyToOne(fetch = FetchType.LAZY)
+  @ManyToOne(fetch = FetchType.EAGER)
```
andricDu, line 77: This doesn't look right and could have big performance implications. 

Why not use the `AnalysisSpecificationBuilder` at query time to dictate which lazy loaded elements you want included?

andricDu (changes requested):
Going from `LAZY` to `EAGER` doesn't look right to me. Happy to discuss.

song-server/src/main/resources/db/migration/V1_7__analysis_state_change_table.sql
```diff
@@ -0,0 +1,24 @@
...
+-- Update Analysis to have a Created At value
+------------------------------------------------------------------
+ALTER TABLE Analysis ADD COLUMN created_at TIMESTAMP NOT NULL DEFAULT NOW();
+ALTER TABLE Analysis ADD COLUMN updated_at TIMESTAMP NOT NULL DEFAULT NOW();
+
+------------------------------------------------------------------
+-- Add state history record for all currently published analysis
+------------------------------------------------------------------
+INSERT INTO analysis_state_change (analysis_id, initial_state, updated_state, updated_at)
+SELECT id, 'UNPUBLISHED', 'PUBLISHED', updated_at FROM analysis WHERE state = 'PUBLISHED';
```
andricDu, line 24: newline

andricDu (changes requested):

song-server/src/main/java/bio/overture/song/server/model/analysis/Analysis.java
joneubank: Thats because I didn't mean to do this. Thanks for catching it.

andricDu: You might want to make your new table also lazy loaded in its join to Analysis.

song-server/src/main/java/bio/overture/song/server/model/analysis/Analysis.java
```diff
@@ -82,6 +85,17 @@
...
 
+  @NotNull
+  @Builder.Default
+  @ToString.Exclude
+  @EqualsAndHashCode.Exclude
+  @OneToMany(
+      mappedBy = ModelAttributeNames.ANALYSIS,
+      cascade = CascadeType.ALL,
+      orphanRemoval = true,
+      fetch = FetchType.EAGER)
```
joneubank, line 96: @andricDu Any issue with this Eager load? It shouldn't have performance issues and is generally relevant info that is being requested in this ticket.

andricDu: `EAGER` is prone to the N+1 problem.

andricDu: For me I would change it to `LAZY` simply because of SONG's usage patterns, a tiny addition in query time gets multiplied a hundred fold. 

I'll let @rtisma decide on this one as he knows the performance pains better than I do.

joneubank: I'm pushing an update that moves this to lazy fetch, so not an issue.

andricDu (approved):
LGTM!

song-server/src/main/java/bio/overture/song/server/model/analysis/Analysis.java
```diff
@@ -69,6 +61,17 @@
...
+  @Column(name = TableAttributeNames.CREATED_AT, nullable = false)
+  @CreationTimestamp
+  private LocalDateTime createdAt;
+
+  @Column(name = TableAttributeNames.UPDATED_AT, nullable = false)
+  @UpdateTimestamp
+  private LocalDateTime updatedAt;
+
+  @Transient private LocalDateTime firstPublishedAt;
+  @Transient private LocalDateTime publishedAt;
```
rtisma, line 73: make sure to add these fields to bio.overture.song.core.model.Analysis (the DTO). This file is used by client, so it can deserialize properly.

song-server/src/main/java/bio/overture/song/server/model/analysis/Analysis.java
```diff
@@ -100,4 +114,23 @@ public AnalysisTypeId getAnalysisType() {
...
   }
+
+  public void populatePublishTimes() {
+    val history = this.analysisStateHistory;
+
+    if (!history.isEmpty()) {
+      val publishHistory =
+          history.stream()
+              .filter(stateChange -> stateChange.getUpdatedState().equals(PUBLISHED.name()))
+              .map(stateChange -> stateChange.getUpdatedAt())
```
rtisma, line 125: this can be simplified to
```java
              .map(AnalysisStateChange::getUpdatedAt)

```

song-server/src/main/java/bio/overture/song/server/model/analysis/Analysis.java
```diff
@@ -100,4 +114,23 @@ public AnalysisTypeId getAnalysisType() {
...
+              .collect(toImmutableList());
```
rtisma, line 126: replace this with
```java
              .collect(toUnmodifiableList());


```
back before java11 we depended on `toImmutableList()` from an external library, but since then java11 add it in.

song-server/src/main/java/bio/overture/song/server/repository/specification/AnalysisSpecificationBuilder.java
```diff
@@ -69,6 +70,9 @@
     if (fetchAnalysisData) {
       root.fetch(ANALYSIS_DATA, LEFT);
     }
+    if (fetchAnalysisStateHistory) {
+      root.fetch(ModelAttributeNames.ANALYSIS_STATE_HISTORY, LEFT);
```
rtisma, line 74: static import

song-server/src/main/resources/db/migration/V1_7__analysis_state_change_table.sql
rtisma, line 24: Add another one for SUPRESSED. We can assume that the initialState was UNPUBLISHED

rtisma (changes requested):
Great work! The tests need to be re-written to use ControllerTests. Their are hibernate annomilies that occur when using the service objects in a test context, that dont exists when making http requests against the local server that running. Serveral other developers (including my self) have noticed feature not working correctly, even though there is a service test. For this reason, as a general rule, for any new functionality we use Controller based Tests.

song-server/src/main/resources/db/migration/V1_7__analysis_state_change_table.sql
rtisma: after discussion, the above requested change is not needed since SUPRESSED is a terminal state

song-server/src/main/java/bio/overture/song/server/model/analysis/Analysis.java
joneubank: I did this, but... This DTO you have defined is not used anywhere, I'm pretty sure. I think adding it to the Controller tests is the first usage of it.

joneubank:
I am pushing commits now that replicate the service tests at the controller level, but there are many complications that make these tests unreliable:
- JSON string output from controller need to be deserialized into DTO objects that are never used in the actual Controller
- Publishing through controller needs to be mocked in the test level in order to run. This has been done before at the Service level but I struggled to reproduce this at the controller level. I would like to revisit mocking this as a dedicated effort later.

I left the controller tests in place, and added a few tests to ensure that the controller is returning the correct Date info (at least enough to deserialize to the DTO version of Analysis).

PR #685, opened 2020-10-20
Add CreatedAt property to AnalysisType/Schema

joneubank:
Note: V8 on the migration file so that it doesnt collide with the V7 in PR: https://github.com/overture-stack/SONG/pull/684


andricDu (approved):
LGTM!

andricDu:
I guess some tests need updates?

song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
```diff
@@ -253,9 +253,14 @@ private AnalysisType commitAnalysisType(
     analysisSchema.setVersion(version);
     log.debug("Registered analysisType '{}' with version '{}'", analysisTypeName, version);
     val resolvedSchemaJson = resolveSchemaJsonView(analysisSchema.getSchema(), false, false);
+
+    val resolvedAnalysisSchema = analysisSchemaRepository.findByNameAndVersion(analysisTypeName, version);
+    val createdAt = resolvedAnalysisSchema.isPresent() ? resolvedAnalysisSchema.get().getCreatedAt() : null;
```
rtisma, line 258: Replace this with
```java
    val createdAt = analysisSchemaRepository
        .findByNameAndVersion(analysisTypeName, version)
        .map(AnalysisSchema::getCreatedAt)
        .orElseThrow(() -> buildServerException(getClass(), ServerErrors.UNKNOWN_ERROR,
            "Cannot find analysisType with name=%s and version=%s", 
             analysisTypeName, version));
```

Where `UNKNOWN_ERROR` maps to a 500 status code, since this is not the client fault.

rtisma (changes requested):

song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
joneubank: This is way better, thanks.

andricDu (approved):

PR #686, opened 2020-10-28
Feat/90 jwt to score

jaserud:
Proposed changes:
- refactor StorageService to not take authorization header and rather take RestTemplate with authorization injected
- Add new profile for client credentials flow to storage/score
- update docker dev setup to reflect new change

Opening PR but with a caveat, see my comment.

### DIFF
--- a/docker/ego-init/init.sql
+++ b/docker/ego-init/init.sql
@@ -295,6 +295,7 @@ ALTER TABLE public.userpermission OWNER TO postgres;
 --
 
 COPY public.egoapplication (name, clientid, clientsecret, redirecturi, description, status, id, type) FROM stdin;
+oauth2-resource	oauth	oauthsecret	http://example.com	song	APPROVED	77f1ef78-7495-4b4a-982a-6b9532dc69fb	CLIENT
 \.
 
 
@@ -344,6 +345,7 @@ COPY public.flyway_schema_history (installed_rank, version, description, type, s
 --
 
 COPY public.groupapplication (group_id, application_id) FROM stdin;
+f2885e96-f74e-4f7a-b935-fb48b18e761d	77f1ef78-7495-4b4a-982a-6b9532dc69fb
 \.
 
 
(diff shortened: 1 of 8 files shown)


docker/ego-init/init.sql
```diff
@@ -295,6 +295,7 @@ ALTER TABLE public.userpermission OWNER TO postgres;
 --
 
 COPY public.egoapplication (name, clientid, clientsecret, redirecturi, description, status, id, type) FROM stdin;
+oauth2-resource	oauth	oauthsecret	http://example.com	song	APPROVED	77f1ef78-7495-4b4a-982a-6b9532dc69fb	CLIENT
```
jaserud, line 298: It looks like the application token comming out of ego has `aud` calims in it 
```
{
  "iat": 1586974587,
  "exp": 1586985387,
  "aud": [
    "ego"
  ],
```
The Spring Oauth2 library in score currently doesn't have an explicit check for this claim so it default to the Spring one. This by default is `oauth2-resource` so any aud claim without this is rejected. Depending on how we want to do things, either score needs to configure expected aud via [ResorceServerConfigurer](https://docs.spring.io/spring-security/oauth/apidocs/org/springframework/security/oauth2/config/annotation/web/configurers/ResourceServerSecurityConfigurer.html#resourceId(java.lang.String)) or maybe ignore the claim or maybe ego needs to make sure this claim isn't their.

jaserud: I've opened the PR anyways since I was able to work around the issue to implement the feature, and its behind a profile so won't work until turned on. But, I'm open to leaving it blocked though if we prefer to wait  until those work is done.

jaserud:
Closing for now. After discussing with team it was agreed this is blocked by ego ticket to remove aud claims.

docker/ego-init/init.sql
jaserud: this is no longer an issue from ego since we removed applcation `aud` claims in https://github.com/overture-stack/ego/issues/513

PR #692, opened 2020-11-04
AnalysisTypePageable checks containing class and update tests

jaserud:
Some background for this PR:
- fix for ticket: https://github.com/overture-stack/score/issues/244
- score client was getting stuck in an infinite loop because the entities endpoint wasn't paging correctly, it's always returning page 0 with size 20 and client keeps requesting until it finds last: https://github.com/overture-stack/score/blob/5d11bb6207ebd6a82d8654236c2f5349965483ba/score-client/src/main/java/bio/overture/score/client/metadata/legacy/LegacyMetadataClient.java#L110
- turns out the AnalysisTypePageable was overwritting the Pageable in the LegacyEntityController with its default values & mapping

Proposed changes: 
- PR updates AnalysisTypePageableResolver to check containing class is AnalysisTypeController
- Also I added tests for LegacyEntitiesController to verify the Pageable is correctly being set

### DIFF
--- a/song-server/src/main/java/bio/overture/song/server/config/WebConfig.java
+++ b/song-server/src/main/java/bio/overture/song/server/config/WebConfig.java
@@ -22,14 +22,12 @@ import bio.overture.song.server.controller.analysisType.AnalysisTypePageableReso
 import java.util.List;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.context.annotation.Profile;
 import org.springframework.web.method.support.HandlerMethodArgumentResolver;
 import org.springframework.web.servlet.config.annotation.ContentNegotiationConfigurer;
 import org.springframework.web.servlet.config.annotation.ViewControllerRegistry;
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
 
 @Configuration
-@Profile("!test")
 public class WebConfig implements WebMvcConfigurer {
 
   @Autowired private SwaggerConfig swaggerConfig;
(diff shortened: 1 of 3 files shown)


rtisma (approved):
LGTM! Make sure to do a sanity test locally via
 https://github.com/overture-stack/genomic-data-playground

Modify the docker-compose.yml  `song-db` service to port-forward 5432 to 9432, so just add
```
ports:
      - 9432:5432
```

Then start the docker services and get them loaded with data
```
make start-storage-services

#Each target when executed creates 4 analyses with 2 files each, so do 5 of them to create 40 files, so that the legacyEntity has 40 entities > DEFAULT_PAGE_SIZE(20)
make test-upload-and-publish
make test-upload-and-publish
make test-upload-and-publish
make test-upload-and-publish
make test-upload-and-publish
make test-upload-and-publish

# Kill song-server so that you can run yours locally via intellig
docker-compose kill song-server 

```
In your intellij, create a run profile with the following envs defined from: https://github.com/overture-stack/genomic-data-playground/blob/master/docker-compose.yml#L148-L167. Since these are using domains that are only internally resolvable, they need to be tweaked to be accessible on the docker host

```
SERVER_PORT: 8080
      SPRING_PROFILES_ACTIVE: "prod,secure,default"
      AUTH_SERVER_URL: http://localhost:9082/o/check_api_key/
      AUTH_SERVER_TOKENNAME: apiKey
      AUTH_SERVER_CLIENTID: song
      AUTH_SERVER_CLIENTSECRET: songsecret
      AUTH_SERVER_SCOPE_STUDY_PREFIX: song.
      AUTH_SERVER_SCOPE_STUDY_SUFFIX: .WRITE
      AUTH_SERVER_SCOPE_SYSTEM: song.WRITE
      SCORE_URL: http://localhost:8087
      SCORE_ACCESSTOKEN: f69b726d-d40f-4261-b105-1ec7e6bf04d5
      MANAGEMENT_SERVER_PORT: 8081
      SPRING_DATASOURCE_USERNAME: postgres
      SPRING_DATASOURCE_PASSWORD: password
      SPRING_DATASOURCE_URL: jdbc:postgresql://localhost:9432/song?stringtype=unspecified
      SPRING_FLYWAY_ENABLED: "true"
      SPRING_FLYWAY_LOCATIONS: "classpath:db/migration"
      ID_USELOCAL: "true"
```

PR #698, opened 2020-11-17
Include whole Analysis in Kafka message payload

joneubank:
- Modify AnalysisServiceSender to read analysis from DB and include in the AnalysisMessage.
- Introduce AnalysisAction to also describe in the kafka message what message was performed (in addition to what state the analysis is now in)

Note: UnsecuredDeepRead now populates schema and data, where previously this was not. This makes some sense for the name, since it is "deep read" but this should be scrutinized for any other impacts it could have to performance. It was necessary for several failing tests that were unable to populate the analysis for the kafka message when these were not read from DB (the JSON serialization threw DB lazy load exceptions on schema/data, even though these are set to JSONIgnore).


andricDu (approved):

song-server/src/test/java/bio/overture/song/server/service/AnalysisServiceSenderTest.java
```diff
@@ -45,66 +53,92 @@
...
-    @NonNull private final AnalysisMessage expectedAnalysisMessage;
+    @NonNull private final String expectedAnalysisMessage;
 
     @Override
     @SneakyThrows
-    public void send(String payload) {
-      val actualAnalysisMessage = JsonUtils.mapper().readValue(payload, AnalysisMessage.class);
+    public void send(String actualAnalysisMessage) {
+      //      val actualAnalysisMessage = JsonUtils.mapper().readValue(payload,
+      // AnalysisMessage.class);
```
rtisma, line 141: delete these unused comments

song-server/src/test/java/bio/overture/song/server/service/AnalysisServiceSenderTest.java
```diff
@@ -45,66 +53,92 @@
...
+  private AnalysisServiceSender createTestAnalysisServiceSender(AnalysisActions action) {
+    val sender = createTestSender(action);
     return new AnalysisServiceSender(SONG_ID, sender, internalAnalysisService);
   }
 
-  private TestSender createTestSender(AnalysisStates state) {
-    return new TestSender(createExpectedAnalysisMessage(state));
+  private TestSender createTestSender(AnalysisActions action) {
+    val analysisMessage = createExpectedAnalysisMessage(action);
+    return new TestSender(JsonUtils.toJson((analysisMessage)));
```
rtisma, line 126: static import `toJson`

song-server/src/main/java/bio/overture/song/server/service/analysis/AnalysisServiceSender.java
```diff
@@ -139,8 +144,9 @@ public Analysis securedDeepRead(@NonNull String studyId, String id) {
   }
 
   private void sendAnalysisMessage(
-      String studyId, String analysisId, AnalysisStates analysisState) {
-    val message = createAnalysisMessage(analysisId, studyId, analysisState, songServerId);
+      String studyId, String analysisId, AnalysisStates analysisState, AnalysisActions action) {
+    val analysis = internalAnalysisService.unsecuredDeepRead(analysisId);
```
rtisma, line 148: this will definetly decrease performance. For every publish/create/supress/unpublish action, there will be double the reading (and therefore double the transactions) since under the hood, `unsecuredDeepRead` is called for each action. 

I think the following partial refactor would eliminate the use of this extra call:
1. modify `AnalysisServiceImpl.checkedUpdateState` to return the analysis object
2. modify the `AnalysisService` interface to return an Analysis object for the publish, unpublish, suppress, and create methods
3. push the "presentation" related stuff (i.e ResponseEntity<String>) that the AnalysisServiceImpl.publish previous returned, to the controller. Likewise with create, unpublish and suppress

This is a bit of work, but will avoid any impact on performance.

song-server/src/test/java/bio/overture/song/server/service/AnalysisServiceSenderTest.java
```diff
@@ -45,66 +53,92 @@
...
+            .analysisData(analysisData)
+            .build();
     this.studyId = RANDOM_GENERATOR.generateRandomAsciiString(15);
   }
 
   @Test
   public void testAnalysisCreate() {
     when(internalAnalysisService.create(studyId, DUMMY_PAYLOAD)).thenReturn(analysisId);
-    val analysisServiceSender = createTestAnalysisServiceSender(UNPUBLISHED);
+    when(internalAnalysisService.unsecuredDeepRead(analysisId)).thenReturn(this.analysis);
```
rtisma, line 86: if the previous comment for the partial refactor is done, then mocking of the unsecuredDeepRead is not needed at all

rtisma (changes requested):

rtisma (changes requested):

rtisma (approved):
LGTM!!

PR #713, opened 2021-03-15
Rc/4.5.0

blabadi:
Make song swagger support path based reverse proxy - as done in ego


blabadi:
will need to address the versions not being updated, tbd

PR #736, opened 2021-11-10
Remove nested null values when parsing Payload

jaserud:
related issue: https://github.com/overture-stack/SONG/issues/737

changes and rationale:
- This PR is adding feature to remove nested null values when parsing a payload
- The reason for doing this is because if we use payload to create an analysis in memory (like [here](https://github.com/overture-stack/SONG/blob/9325cf61a05e2461242235418ea31364392fd71d/song-server/src/main/java/bio/overture/song/server/service/analysis/AnalysisServiceImpl.java#L133)), that analysis could have dynamic fields that are null. However, when we write to db these are not being saved. This leads to inconsistency since analysis in memory has fields that are null, but db save removed them. 
- So, by removing any null fields when we parse the Payload from the submit function, any one using it won't have to deal with the inconsistency

### DIFF
--- a/song-server/src/main/java/bio/overture/song/server/model/dto/Payload.java
+++ b/song-server/src/main/java/bio/overture/song/server/model/dto/Payload.java
@@ -2,16 +2,15 @@ package bio.overture.song.server.model.dto;
 
 import bio.overture.song.core.model.AnalysisTypeId;
 import bio.overture.song.core.model.DynamicData;
+import bio.overture.song.core.utils.JsonUtils;
 import bio.overture.song.server.model.entity.FileEntity;
 import bio.overture.song.server.model.entity.composites.CompositeEntity;
 import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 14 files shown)


song-server/src/main/java/bio/overture/song/server/model/dto/Payload.java
```diff
@@ -21,9 +20,31 @@
...
   private AnalysisTypeId analysisType;
   private List<CompositeEntity> samples;
   private List<FileEntity> files;
+
+  @SneakyThrows
+  public static Payload parse(JsonNode jsonStr) {
+    // convert to hashMap
+    val jsonNode = MAPPER.readValue(jsonStr.toString(), HashMap.class);
+    // writeValueAsString will remove null/empty values from map, including nested
+    val sanitized = MAPPER.writeValueAsString(jsonNode);
```
Buwujiu, line 35: The reason `writeValueAsString` will remove null/empty values is because the `MAPPER`  includes non-null non-empty values as defined in `payloadObjectMapper` method?

jaserud: Yes, that is correct. I guess this comment can be more specific abotu that.

PR #738, opened 2021-11-11
new getAnalysis endpoint with pagination

Buwujiu:
This pr adds a new `getAnalysis` endpoint that has pagination. 
- no change made to the old `getAnalysis`, we can migrate to the new paginated endpoint after it's tested.
- The paginated endpoint takes a `page` and a `size` query params.

### DIFF
--- a/song-server/src/main/java/bio/overture/song/server/service/analysis/AnalysisServiceImpl.java
+++ b/song-server/src/main/java/bio/overture/song/server/service/analysis/AnalysisServiceImpl.java
@@ -38,7 +38,9 @@ import static bio.overture.song.core.utils.JsonUtils.fromJson;
 import static bio.overture.song.core.utils.JsonUtils.readTree;
 import static bio.overture.song.core.utils.JsonUtils.toJsonNode;
 import static bio.overture.song.core.utils.Separators.COMMA;
-import static bio.overture.song.server.model.enums.ModelAttributeNames.ANALYSIS_TYPE;
+import static bio.overture.song.server.model.enums.ModelAttributeNames.*;
+import static bio.overture.song.server.repository.specification.AnalysisSpecificationBuilder.equalsStudyPredicate;
+import static bio.overture.song.server.repository.specification.AnalysisSpecificationBuilder.whereStatesInPredicate;
 import static bio.overture.song.server.utils.JsonSchemas.PROPERTIES;
 import static bio.overture.song.server.utils.JsonSchemas.REQUIRED;
 import static bio.overture.song.server.utils.JsonSchemas.buildSchema;
@@ -47,6 +49,7 @@ import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static java.lang.String.format;
 import static java.util.Objects.isNull;
+import static javax.persistence.criteria.JoinType.LEFT;
 
 import bio.overture.song.core.model.AnalysisTypeId;
 import bio.overture.song.core.model.enums.AnalysisStates;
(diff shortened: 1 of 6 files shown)


song-server/src/main/java/bio/overture/song/server/service/analysis/AnalysisServiceImpl.java
```diff
@@ -208,6 +218,63 @@ public void updateAnalysis(
...
+  @Override
+  public GetAnalysisResponse getAnalysis(
+      @NonNull String studyId,
+      @NonNull Set<String> analysisStates,
+      @NonNull int page,
+      @NonNull int size) {
+    studyService.checkStudyExist(studyId);
+    resolveSelectedAnalysisStates(analysisStates);
+
+    val sort = new Sort(Sort.Direction.ASC, ANALYSIS_ID);
```
jaserud, line 230: What do you think about making the sort a requestParam too? Like have a set for `fieldNames` to sort on and the `order` to chose asc or desc?? If we prefer/require only a default sort, maybe something like `publishedAt` or `createdAt` makes more sense??

song-server/src/main/java/bio/overture/song/server/service/analysis/AnalysisServiceImpl.java
```diff
@@ -208,6 +218,63 @@ public void updateAnalysis(
...
+        };
+
+    val analyses = repository.findAll(spec, pageRequest);
+
+    analyses.forEach(
+        a -> {
+          val id = a.getAnalysisId();
+          a.setFiles(unsecuredReadFiles(id));
+          a.setSamples(readSamples(id));
+          a.populatePublishTimes();
```
jaserud, line 259: Just want to double check, it looks like this pr is just to add paging not to fix the in memory joins?

Buwujiu: just adding paging, there is another ticket for investigating how to improve performance.

song-server/src/main/java/bio/overture/song/server/service/analysis/AnalysisServiceImpl.java
Buwujiu: I checked a bunch of analyses and there are quite a few of them (same program)that are created at the same time, so sorting by `createdAt` doesn't provide much functionality. Not sure about `publishedAt`.

jaserud: Hmm ok, ids are always unique so order_by/sort on them will always return the same result. I suppose if we just want the default sort with no option to sort via params, analysisId (with the studyId we already get) is good enough.

song-server/src/main/java/bio/overture/song/server/service/analysis/AnalysisServiceImpl.java
```diff
@@ -208,6 +218,63 @@ public void updateAnalysis(
...
+      @NonNull int page,
+      @NonNull int size) {
+    studyService.checkStudyExist(studyId);
+    resolveSelectedAnalysisStates(analysisStates);
+
+    val sort = new Sort(Sort.Direction.ASC, ANALYSIS_ID);
+    val pageRequest = PageRequest.of(page, size, sort);
+
+    val spec =
+        new Specification<Analysis>() {
```
jaserud, line 234: I'm curious why is this not a part of the `AnalysisSpecificationBuilder` class :stuck_out_tongue:

Buwujiu: I suppose I could move the logic to `AnalysisSpecificationBuilder`

jaserud (approved):
LGTM

Buwujiu:
This PR is merged, but is overwritten by https://github.com/overture-stack/SONG/pull/743 as the new PR has the latest implementation. No need to test this ticket.

PR #743, opened 2022-01-05
add a new paginated get analysis endpoint with improved perf

Buwujiu:
PR for song getAnalysis endpoint performance upgrade. 

This PR should significantly reduce the time needed to fetch analyses for studies. For example, local performance tests show a result of 2.3 seconds.
New implementation of old `getAnalysis` endpoint, this PR adds a new `paginated` getAnalysis endpoint with `limit` and `offset` parameters to specify paged analysis. 
- adds 4 sql migration scripts to create 4 postgres functions which fetch analysis from DB. When migration completes, db should have these functions available for song service to use. 
- adds the following java classes to map function returns: `DataEntity.java`, `AnalysisDataJoin.java`, `AnalysisSchemaJoin.java`, `AnalysisStateChangeJoin.java`.
- adds a new `UpgradedAnalysisRepository.java` to call DB functions for getting analysis data.


song-server/src/main/java/bio/overture/song/server/service/analysis/AnalysisServiceImpl.java
```diff
@@ -214,37 +201,69 @@ public void updateAnalysis(
   public GetAnalysisResponse getAnalysis(
```
Buwujiu, line 201: the new service method for paginated endpoint

song-server/pom.xml
```diff
@@ -160,6 +160,11 @@
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.vavr</groupId>
```
blabadi, line 164: where is this used ?

blabadi (approved):

song-server/pom.xml
Buwujiu: AnalysisServiceImpl.java

Buwujiu: https://github.com/overture-stack/SONG/pull/743/files#diff-3ad7aede999b78dca2c87366270896b4684688e893d4ba6168d1951ad06d02c3R232

PR #745, opened 2022-01-07
Add fileType TXT to base schema and file_type enum

joneubank:
Migration based on migration v1_4 which added CRAM and CRAI to the same enum. Since that worked previously this migration uses the same process copy and pasted with value updates.


joneubank:
the JSON files were only changed to add the "TXT" file type, all the other updates are auto formatting.

song-server/src/main/resources/db/migration/V1_9__filetype_enum_add_txt.sql
```diff
@@ -0,0 +1,20 @@
+ALTER TYPE file_type RENAME TO _file_type;
```
blabadi, line 1: I'm not very familiar with enums, but I thought it would just be something like this: 
`ALTER TYPE enum_name ADD VALUE 'test4' AFTER 'test3';`

blabadi (approved):

song-server/src/main/resources/db/migration/V1_9__filetype_enum_add_txt.sql
joneubank: yeah I looked up the same and had prepared that, but in migration v1_4 it was done this way so I thought to stay consistent with what worked previously...

PR #747, opened 2022-01-12
Release Candidate 4.8.0

Buwujiu:
Release Note:
- Song performance upgrades:  a new paginated `getAnalysis` endpoint to replace legacy endpoint. 
- Add fileType TXT to base schema and file_type enum.
- Create, send analysis after transaction commit.

### DIFF
--- a/Dockerfile.icgc25k
+++ b/Dockerfile.icgc25k
@@ -1,7 +1,7 @@
 ###############################
 # Song Server for ICGC25k
 ###############################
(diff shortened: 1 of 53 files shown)


blabadi:
adding @joneubank since he is also working on SONG now, @joneubank does this affect you or do you know any reason not to merge this ?

joneubank:
Merging is fine, but once we resolve the dependency issue that jenkins ran into.

PR #755, opened 2022-03-01
remove temp table from get_analysis function db script

Buwujiu:


song-server/src/main/resources/db/migration/V1_10__create_function_get_analysis.sql
```diff
@@ -24,30 +24,8 @@ returns table (
 ) as
 $$
 	begin
-		 -- step A: analysis_file_join is the result of joining Analysis table with File table, and getting file info from Info table
```
blabadi, line 27: you cannot modify existing scripts, you need to add new script since this one is already executed

Buwujiu: added a new script

blabadi (approved):

PR #762, opened 2022-04-04
Upgrade SpringBoot to 2.6.6

joneubank:
In response to Spring4Shell exploit, upgrading to latest SpringBoot 2.6.6
This required updating the versions of:
* spring-cloud to 2021.0.1
* jackson to 2.13.2

Upgrade is mostly seemless but for 3 small changes needed to build:
* Pageable has an additional method to implement `withPage` - like other methods that return a Pageable object this just returns null for our use case
* Sort has no public constructors... switching to `Sort.by()`
* `AutoConfigureMockMvc` no longer has a `secure` property to set

Additional Changes found through testing:
* `application.yaml` profile structure is now `spring.config.activate.on-profile` instead of `spring.profiles`
* Spring Profiles cannot specify other Profiles to be used anymore, so the cases of `spring.profile.include` had to be removed. Those were:
    * `test` included `dev`
    * `dev` included `fastTest`
    * `jwt` included `secure`
* Spring no longer uses `bootstrap.properties`, the bootstrap value to disable vault was moved to `application.yaml`
* `SongPrettyPrinter` Implementation of `DefaultPrettyPrinter` needs to override `createInstance()`

## Notes when Upgrading:
* If using `jwt` profile, be sure to also use `secure` profile

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -514,7 +514,6 @@
         <everit.jsonschema.version>1.11.1</everit.jsonschema.version>
         <plugin.fmt.version>2.9</plugin.fmt.version>
         <plugin.maven.dependency.version>3.1.1</plugin.maven.dependency.version>
-        <maven.dependency.analyzer.version>1.12.0</maven.dependency.analyzer.version>
     </properties>
 
 </project>


blabadi:
when I worked on ego, the following changed: 
- spring cloud version had to be compatible with boot version, for 2.6.6 it's 2021.x.x: https://spring.io/projects/spring-cloud 
- spring boot changed the way profiles work : https://github.com/overture-stack/ego/pull/615/files#diff-7bc26be01d3e7c9c566f2a28dc1b5cd87dbdd5cb619184f46eb2bf223bf7825dR203

now for runtime configuration changes (won't affect this PR as far as I know) :
- spring boot removed support for bootstrap.yaml / properties, had to do this: https://github.com/OICR/argo-infra/blob/master/helm/ego/qa/values.yaml#L58
- vault values had to change in configurations: https://wiki.oicr.on.ca/display/OV/Ego+Spring+Oauth2+Update+-+5.x.x#EgoSpringOauth2Update5.x.x-Springbootupgrades(5.0.0)-5points

pom.xml
```diff
@@ -529,7 +514,7 @@
         <everit.jsonschema.version>1.11.1</everit.jsonschema.version>
         <plugin.fmt.version>2.9</plugin.fmt.version>
         <plugin.maven.dependency.version>3.1.1</plugin.maven.dependency.version>
-        <maven.dependency.analyzer.version>1.11.1</maven.dependency.analyzer.version>
+        <maven.dependency.analyzer.version>1.12.0</maven.dependency.analyzer.version>
```
blabadi, line 517: is this still needed with the plugin removed ?

blabadi (approved):

PR #767, opened 2022-06-07
add new enum value to specimen_type

Buwujiu:


joneubank (approved):
Changes look fine, look like they follow the established pattern (which we should review later).

Note that there are issues running Song with security enabled since the Spring upgrade, so don't merge this to master yet. I need to follow up on those errors. This will fail to start in the overture QA env but not because of your changes.

blabadi (approved):

Buwujiu:
Thanks for the headsup @joneubank !

PR #768, opened 2022-06-11
Fix Secure Profiles after Spring Upgrade

joneubank:
The SpringBoot upgrade broke Spring WebSecurity which only became an issue when running secure and jwt profiles (which are used in all production cases).

The `secure` profile continues to enable the WebSecurity configuration which requires an ApiKey to access protected resources. This now configures an Oauth2ResourceServer using a custom OpaqueToken introspector that works with Ego.
Note: The Oauth2 token introspection response from ego is missing the `active` property that by default indicates if the token is currently valid.

The `jwt` profile changes Oauth2ResourceServer to accept both ApiKey and JWT as the Bearer token. The JWT validation uses the out of the box Spring beans since the jwt public key path and valdiation work as expected by the Oauth standard.

Additional updates are made to extract the scopes from both the ApiKey response and the JWT, depending on which is provided to the protected endpoint pre-authorization method.


song-server/src/main/java/bio/overture/song/server/config/JwtSecurityConfig.java
```diff
@@ -0,0 +1,87 @@
...
+import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationProvider;
+import org.springframework.security.oauth2.server.resource.authentication.OpaqueTokenAuthenticationProvider;
+import org.springframework.security.oauth2.server.resource.introspection.OpaqueTokenIntrospector;
+import org.springframework.stereotype.Component;
+
+@Slf4j
+@Component
+@Getter
+@Setter
+@Profile("jwt")
```
blabadi, line 47: this profile won't work without the `secure` profile because it uses the `OpaqueTokenIntrospector` if so why not merge them together instead of having different profiles ?

song-server/src/main/java/bio/overture/song/server/utils/Scopes.java
```diff
@@ -34,7 +43,29 @@ public static long extractExpiry(Map<String, ?> map) {
     return 0L;
   }
 
-  private static Set<String> getScopes(OAuth2Authentication o2auth) {
-    return o2auth.getOAuth2Request().getScope();
+  private static Set<String> getJwtScopes(JwtAuthenticationToken jwt) {
+    Set<String> output = new HashSet();
+    try {
+      val context = jwt.getToken().getClaim("context");
```
blabadi, line 49: can context be null here ?

song-server/src/main/java/bio/overture/song/server/config/JwtSecurityConfig.java
```diff
@@ -0,0 +1,87 @@
...
+  @SneakyThrows
+  public void configure(HttpSecurity http) {
+    http.oauth2ResourceServer(
+        oauth2 -> oauth2.authenticationManagerResolver(this.tokenAuthenticationManagerResolver()));
+  }
+
+  private boolean useJwt(HttpServletRequest request) {
+    val authorizationHeaderValue = request.getHeader(HttpHeaders.AUTHORIZATION);
+
+    if (authorizationHeaderValue != null && authorizationHeaderValue.startsWith("Bearer")) {
```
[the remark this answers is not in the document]
joneubank: Is this something we have in the config? Bearer is standard string here and I believe the SpringSecurity OAuth2 code expects it... So I think it is safe to expect it. i'll look for a constant in the code instead of using the string litteral.

song-server/src/main/java/bio/overture/song/server/config/JwtSecurityConfig.java
```diff
@@ -0,0 +1,87 @@
...
+      String token = authorizationHeaderValue.substring(7, authorizationHeaderValue.length());
```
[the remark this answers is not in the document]
joneubank: Thank you. I'll use that.

song-server/src/main/java/bio/overture/song/server/config/JwtSecurityConfig.java
joneubank: I agree we should just use it by default. I was keeping it for reasons of "thats how it was done before" but I don't know why you wouldn't enable JWT.

song-server/src/main/java/bio/overture/song/server/utils/Scopes.java
joneubank: instanceof returns false when context is null. should work as expected.

song-server/src/main/java/bio/overture/song/server/config/JwtSecurityConfig.java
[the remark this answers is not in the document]
joneubank: The only available constant provided by spring security is in a deprecated class so I left it as is. The default JwtAuthenticationProvider expects this anyways so it should be correct.

joneubank:
I have removed the JWT profile, `secure` profile now handles both Ego ApiKey and JWTs.

blabadi (approved):

b-f-chan:
@Buwujiu - Can this be moved to Q?  PR is approved

Issue #775, opened 2022-08-04
Add new enum `Tumour - unknown if derived from primary or metastatic` to specimen_type


Buwujiu:
Original request ticket is https://github.com/icgc-argo/argo-dictionary/issues/340#issuecomment-1090829348.
Based on Hardeep's comment: https://github.com/icgc-argo/argo-dictionary/issues/340#issuecomment-1090829348, we need to add `Tumour - unknown if derived from primary or metastatic` to specimen_type, NOT `Malignant neoplasm without specification of site`. However, as I was not aware of the comment,  `Malignant neoplasm without specification of site` was added to `specimen_type` in song, we need to:
1. Remove `Malignant neoplasm without specification of site` from `specimen_type`.
2. Add  `Tumour - unknown if derived from primary or metastatic` to `specimen_type`.

Buwujiu:
Edmund tested in prod:
work for submisison.song and song :
- https://submission-song.rdpc.cancercollaboratory.org/studies/TEST-INTL/analysis/e380bf5a-5cb6-438a-80bf-5a5cb6938a70
- https://song.rdpc.cancercollaboratory.org/studies/TEST-INTL/analysis/951a3478-7d9c-4cd5-9a34-787d9c6cd5ad
- clinical rule not in place thus cannot test in clinical.song

PR #778, opened 2022-09-13
replaced the enum value in specimen type for ticket 'Add new enum #775'

UmmulkiramR:
“Malignant neoplasm without specification of site” is a newly added value. 
Ticket #775 is about replacing this value with a new value "Tumour - unknown if derived from primary or metastatic"
As a part of this change a migration script V1_17__specimen_type_update_enum.sql is prepared to update the specimen_type enum. 
This enum is referenced in the specimen table but this enum value being newly added is not associated with any records.


song-server/src/main/resources/db/migration/V1_16__specimen_type_update_enum.sql
```diff
@@ -0,0 +1 @@
+ALTER TYPE specimen_type RENAME VALUE 'Malignant neoplasm without specification of site' TO 'Tumour - unknown if derived from primary or metastatic'
```
Buwujiu, line 1: There is already a V16 script exists in song, please rename it to V17.

Buwujiu (changes requested):

song-server/src/main/resources/db/migration/V1_16__specimen_type_update_enum.sql
UmmulkiramR: Filename changed.

joneubank (approved):

PR #780, opened 2022-09-28
Add analysisId as key to kafka messages

joneubank:
When sending messages to a kafka topic with multiple partitions, we want to ensure that messages for each analysis end up in the same partition. This ensures that the consuming service can process them in chronological order. To accomplish this, we add the analysisId as the key to the kafka message.

To illustrate this scenario, imagine the `song_analysis` topic has 2 partitions to support 2 instances of **Maestro**. Now, a user creates a new analysis in **Song** which generates a kafka message with the `analysisState` = `UNPUBLISHED`. Then the user publishes that analysis, generating a kafka message with the `analysisState` = `PUBLISHED`. If those two messages end up in different partitions, it is possible that the `PUBLISHED` event is processed in advance of the `UNPUBLISHED` event, causing the analysis to be removed from the **Elasticsearch** index. If both messages are guaranteed to be placed in the same partition, then they will be consumed by the same instance of **Maestro**, processed in order, and this race condition won't occur.

### DIFF
--- a/song-server/src/main/java/bio/overture/song/server/service/analysis/AnalysisServiceSender.java
+++ b/song-server/src/main/java/bio/overture/song/server/service/analysis/AnalysisServiceSender.java
@@ -1,12 +1,7 @@
 package bio.overture.song.server.service.analysis;
 
-import static bio.overture.song.core.model.enums.AnalysisActions.CREATE;
-import static bio.overture.song.core.model.enums.AnalysisActions.PUBLISH;
-import static bio.overture.song.core.model.enums.AnalysisActions.SUPPRESS;
-import static bio.overture.song.core.model.enums.AnalysisActions.UNPUBLISH;
-import static bio.overture.song.core.model.enums.AnalysisStates.PUBLISHED;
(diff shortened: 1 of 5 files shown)


song-server/src/main/java/bio/overture/song/server/service/analysis/AnalysisServiceSender.java
```diff
@@ -174,6 +169,6 @@ public Analysis securedDeepRead(@NonNull String studyId, String id) {
   private void sendAnalysisMessage(
       Analysis analysis, AnalysisStates analysisState, AnalysisActions action) {
     val message = createAnalysisMessage(action, analysis, songServerId);
-    sender.send(toJson(message));
+    sender.send(toJson(message), message.getAnalysisId());
```
joneubank, line 172: Only place in the code base where kafka messages are sent. We specify the key to be the AnalysisID from the message.

lrivera-oicr: looks good to me 👍

PR #786, opened 2022-10-17
deep read analysis

lrivera-oicr:
Fetch the Analysis with all Files, Samples and info on PUT and PATCH analysis operations, this is required as Maestro needs to index the complete analysis object.


song-server/src/main/java/bio/overture/song/server/service/analysis/AnalysisServiceImpl.java
```diff
@@ -156,8 +156,8 @@ public Analysis updateAnalysis(
     // Validate the updateAnalysisRequest against the scheme
     validateUpdateRequest(updateAnalysisRequest, newAnalysisSchema);
 
-    // Now that the request is validated, fetch the old analysis
-    val analysis = get(analysisId, true, true, true);
+    // Now that the request is validated, is safe to fetch the old analysis with all files, samples and info
```
caravinci, line 159:
```suggestion
    // Now that the request is validated, it's safe to fetch the old analysis with all files, samples and info
```

joneubank (approved):

PR #788, opened 2022-11-17
release 5.0.3

UmmulkiramR:
Changed the version number for the release.


joneubank (changes requested):
Need to update the version in all project POMs, not just the root. You can see the others that need updates in the file diff.

joneubank (approved):

PR #797, opened 2023-03-02
Fix song docker demo

leoraba:
- Fix broken Unit tests by JSON parsing.
- Use a multiplatform JDK build image on Dockerfile
- Adding missing environment variables used by docker composer


caravinci (approved):
LGTM 👍

PR #841, opened 2024-07-02
Keycloak integration

leoraba:
Integrate Keycloak as Auth Server Epic #832


docker-compose.yml
```diff
@@ -86,7 +86,7 @@ services:
       - "./docker/scratch/score-server-logs:/score-server/logs"
 
   score-client:
-    image: overture/score:5.0.0
+    image: overture/score:5.10.0
```
joneubank, line 89: Should this just be `latest` or `edge`?

joneubank (approved):

docker-compose.yml
leoraba: I have changed this to `latest` if we're looking for the latest stable version.

joneubank:
Creating ticket to track need for documenting the auth provider options: #842

Issue #848, opened 2024-08-21
Rename ICGC Bucket Names to score.data and score.state in Docker-Compose and Makefile


Azher2Ali:
## Summary of request
The docker-compose setup provides default bucket names that reference ICGC, this now will be renamed these for score.

Change default bucket names to score.data and score.state

     BUCKET_NAME_OBJECT: oicr.icgc.test
     BUCKET_NAME_STATE: oicr.icgc.test

Also, Updates to the Make file to reference the new bucket names. There are several lines in this section that reference `oicr.icgc.test` that should all be replaced with` score.data`

## Additional context
Include any relevant context or screenshots regarding your request here.

https://github.com/overture-stack/SONG/blob/develop/docker-compose.yml#L57

https://github.com/overture-stack/SONG/blob/develop/docker-compose.yml#L58

https://github.com/overture-stack/SONG/blob/develop/docker-compose.yml#L132

https://github.com/overture-stack/SONG/blob/develop/dev/docker-compose.yml#L38

https://github.com/overture-stack/SONG/blob/develop/dev/docker-compose.yml#L39

https://github.com/overture-stack/SONG/blob/develop/Makefile#L93

https://github.com/overture-stack/SONG/blob/develop/Makefile#L94

https://github.com/overture-stack/SONG/blob/develop/Makefile#L97

https://github.com/overture-stack/SONG/blob/develop/Makefile#L99

https://github.com/overture-stack/SONG/blob/develop/Makefile#L102

https://github.com/overture-stack/SONG/blob/develop/Makefile#L103

https://github.com/overture-stack/SONG/blob/develop/Makefile#L104

Issue #846, opened 2024-08-19
Remove ICGC Object Storage Configuration from Codebase


Azher2Ali:
## Summary of request
This request involves the removal of the docker/object-storage-init/data/oicr.icgc.test directory from the codebase. The directory contains object storage configurations related to the ICGC project, which are no longer relevant or in use. Removing these obsolete configurations is part of the effort to clean up the codebase and eliminate outdated ICGC references.
 
## Desired solution
The desired solution is to delete the docker/object-storage-init/data/oicr.icgc.test directory and any related references in the codebase. This will help streamline the project by removing unnecessary files and reducing potential confusion regarding unused configurations.

## Additional context
Include any relevant context or screenshots regarding your request here.
https://github.com/overture-stack/SONG/tree/develop/docker/object-storage-init/data/oicr.icgc.test

https://github.com/overture-stack/SONG/blob/develop/Makefile#L93

https://github.com/overture-stack/SONG/blob/develop/Makefile#L97

https://github.com/overture-stack/SONG/blob/develop/Makefile#L99

https://github.com/overture-stack/SONG/blob/develop/Makefile#L102

https://github.com/overture-stack/SONG/blob/develop/Makefile#L103

https://github.com/overture-stack/SONG/blob/develop/Makefile#L104

joneubank:
This data is used in the docker-compose setup which is in place for dev purposes. It is mounted here: https://github.com/overture-stack/SONG/blob/develop/docker-compose.yml#L132

We need to instead update the default bucket names in the docker-compose so they don't reference `icgc` and then update the directory names for the data that you pointed to so that it uses the new data bucket name.

PR #849, opened 2024-08-21
Refactor ICGC Bucket Names to SCORe in Docker-Compose and Makefile

Azher2Ali:
This PR renames the default bucket names in the Docker-Compose setup from oicr.icgc.test to score.data and score.state to align with the SCORe project. It also updates the Makefile to reference the new bucket names. All instances of oicr.icgc.test in the relevant sections have been replaced with score.data to remove outdated ICGC references and standardize the configuration. #848 #846

### DIFF
--- a/dev/docker-compose.yml
+++ b/dev/docker-compose.yml
@@ -35,8 +35,8 @@ services:
       AUTH_SERVER_UPLOADSCOPE: song.upload
       SERVER_URL: http://localhost:8087
       SERVER_PORT: 8087
-      BUCKET_NAME_OBJECT: oicr.icgc.test
-      BUCKET_NAME_STATE: oicr.icgc.test
+      BUCKET_NAME_OBJECT: score.data
+      BUCKET_NAME_STATE: score.test
       COLLABORATORY_DATA_DIRECTORY: data
       OBJECT_SENTINEL: heliograph
       METADATA_URL: "http://localhost:8080"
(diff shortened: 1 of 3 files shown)


dev/docker-compose.yml
```diff
@@ -35,8 +35,8 @@ services:
       AUTH_SERVER_UPLOADSCOPE: song.upload
       SERVER_URL: http://localhost:8087
       SERVER_PORT: 8087
-      BUCKET_NAME_OBJECT: oicr.icgc.test
-      BUCKET_NAME_STATE: oicr.icgc.test
+      BUCKET_NAME_OBJECT: score.data
+      BUCKET_NAME_STATE: score.test
```
joneubank: can we rename this to `score.state`?

joneubank (changes requested):

dev/docker-compose.yml
joneubank: Looks like in score we made this `score.test`, missed this detail earlier. I'll update that to make consistent.

joneubank: Cancel the above notes... There was a reason they used the same bucket name for both of these buckets... there is mounted directory for the test data and we renamed the bucket to `score.data`, we would either need to create a directory for the state bucket or reuse the data bucket name.

Therefore, I think its easiest to keep both bucket names the same for the local docker compose setup. Let's update this to `BUCKET_NAME_STATE: score.data`.

joneubank (approved):

PR #854, opened 2024-08-29
Enable Dynamic Definition of File Types for Analysis Types in Base Schema

Azher2Ali:
This PR introduces changes to the base schema and dynamic schemas to support the flexible definition of file types for different analysis types. Key modifications include:

**Base Schema Updates:**

- [x] fileType is now defined as its own separate definition.

- [x] The only validation on fileType is that it must be a string.

- [x] file.fileType now references the fileType definition.

- [x] fileType remains a required property for file.

**Analysis Type Registration Schema:**

Allows dynamic schemas to define additional restrictions on file.fileType.
If the user provides a fileType definition during analysis type registration, it will merge with the base schema definition, allowing for analysis-specific file type validation./

This update also includes a data migration to refine the list of valid file types associated with the legacy analysis types (sequencing-experiment and variant-calling) that were pre-loaded into the Song database.

**Summary of Changes:**

**Specific File Type Lists:**
The list of valid file types for sequencing-experiment and variant-calling analysis types has been updated to be specific to these analysis types, replacing the previous implementation where a broad list of file types was allowed.
This change ensures that only relevant file types can be used with each analysis type, enhancing data validation and integrity.


song-server/src/main/resources/schemas/analysis/analysisTypeRegistration.json
```diff
@@ -10,9 +10,25 @@
...
+                                { "$ref": "classpath:/schemas/analysis/analysisBase.json#/definitions/file/fileType"},
+                                {
+                                    "type": "string",
+                                    "enum": ["VCF", "BAM", "OTHER_CUSTOM_TYPE"]
+                                }
+                            ]
+                        }
+                    }
+                }
+            }
```
joneubank, line 29: It looks like, you have provided these `enum` values as an example... This schema is used to validate the content submitted when a user is defining a dynamic schema, and this would allow them to provide a schema with `file.fileType: "OTHER_CUSTOM_TYPE"`. This is definitely not what we want.

What we want is the user to provide additional restrictions to the `classpath:/schemas/analysis/analysisBase.json#/definitions/file/fileType` definition, for instance providing a custom enum value. This is so that when a new analysis is submitted those validation rules are applied.

joneubank (changes requested):

PR #855, opened 2024-09-06
Dynamic Schema Support for Analysis-Specific File Types and Adding data migration

Azher2Ali:
This PR introduces changes to the base schema and dynamic schemas to support the flexible definition of file types for different analysis types. Key modifications include:

Base Schema Updates:

 fileType is now defined as its own separate definition.

 The only validation on fileType is that it must be a string.

 file.fileType now references the fileType definition.

 fileType remains a required property for file.

Analysis Type Registration Schema:

Allows dynamic schemas to define additional restrictions on file.fileType.
If the user provides a fileType definition during analysis type registration, it will merge with the base schema definition, allowing for analysis-specific file type validation./

This update also includes a data migration to refine the list of valid file types associated with the legacy analysis types (sequencing-experiment and variant-calling) that were pre-loaded into the Song database.

Summary of Changes:

Specific File Type Lists:
The list of valid file types for sequencing-experiment and variant-calling analysis types has been updated to be specific to these analysis types, replacing the previous implementation where a broad list of file types was allowed.
This change ensures that only relevant file types can be used with each analysis type, enhancing data validation and integrity.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,9 +57,12 @@
             <url>https://jcenter.bintray.com/</url>
         </repository>
         <repository>
-            <id>spring-releases</id>
+            <id>lib-releases</id>
             <name>Spring Releases</name>
             <url>https://repo.spring.io/libs-release</url>
+            <snapshots>
+                <enabled>false</enabled>
+            </snapshots>
         </repository>
     </repositories>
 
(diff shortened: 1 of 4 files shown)


joneubank:
Is this now a duplicate of https://github.com/overture-stack/SONG/pull/858 ?

Can we close this?

PR #858, opened 2024-09-19
Enable Dynamic Schema Support for File Type Definitions in Analysis Types

Azher2Ali:
1. **Base Schema Updates**:
   - `fileType` is now its own definition, allowing any string value.
   - `file.fileType` references this new definition, ensuring that it is always a required property for files.
   
2. **Analysis Type Registration Schema**:
   - If a user provides a `file.fileType` definition, it merges with the base schema definition.
   - Users can now add additional restrictions on `file.fileType` within the dynamic schema.
   
3. **Data Structure Enhancements**:
   - Added support for an optional `options.fileTypes` array in the `analysisType` dynamic schema, allowing users to define specific allowed file types.
   - Updated the controller and `analysisTypeService.register` to accept the `options` object.

4. **Validation Enforcement**:
   - The analysis submission/update process has been updated to enforce allowed `fileTypes` from the schema, in addition to JSON Schema validation.
   - If no file types are defined in the dynamic schema, all file types are accepted.

### Notes:
- Additional validation logic has been implemented to ensure that the file types of each file submitted in an analysis match the allowed types defined in the dynamic schema, if provided.
- If no specific file types are defined for an analysis type, the check is skipped.

### Impact:
This change provides a more flexible and dynamic way to handle file type validations for various analysis types, paving the way for future enhancements and custom analysis schemas.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,9 +57,12 @@
             <url>https://jcenter.bintray.com/</url>
         </repository>
         <repository>
-            <id>spring-releases</id>
+            <id>lib-releases</id>
             <name>Spring Releases</name>
             <url>https://repo.spring.io/libs-release</url>
+            <snapshots>
+                <enabled>false</enabled>
+            </snapshots>
         </repository>
     </repositories>
 
--- a/pom.xml
+++ b/pom.xml
@@ -56,14 +56,14 @@
             <name>jcenter</name>
             <url>https://jcenter.bintray.com/</url>
         </repository>
-        <repository>
-            <id>lib-releases</id>
-            <name>Spring Releases</name>
-            <url>https://repo.spring.io/libs-release</url>
-            <snapshots>
-                <enabled>false</enabled>
-            </snapshots>
-        </repository>
+       <repository>
+          <id>lib-releases</id>
+          <name>Spring Lib Releases</name>
+          <url>https://repo.spring.io/libs-release</url>
+          <snapshots>
+              <enabled>false</enabled>
+          </snapshots>
+      </repository>
     </repositories>
 
     <dependencyManagement>
--- a/pom.xml
+++ b/pom.xml
@@ -56,14 +56,14 @@
             <name>jcenter</name>
             <url>https://jcenter.bintray.com/</url>
         </repository>
-       <repository>
-          <id>lib-releases</id>
-          <name>Spring Lib Releases</name>
-          <url>https://repo.spring.io/libs-release</url>
-          <snapshots>
-              <enabled>false</enabled>
-          </snapshots>
-      </repository>
+        <repository>
+            <id>lib-releases</id>
+            <name>Spring Releases</name>
+            <url>https://repo.spring.io/libs-release</url>
+            <snapshots>
+                <enabled>false</enabled>
+            </snapshots>
+        </repository>
     </repositories>
 
     <dependencyManagement>
--- a/song-core/src/main/java/bio/overture/song/core/model/Options.java
+++ b/song-core/src/main/java/bio/overture/song/core/model/Options.java
@@ -1,0 +1,12 @@
+package bio.overture.song.core.model;
+
+import java.util.List;
+import lombok.*;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class Options {
+  private List<String> fileTypes;
+}
--- a/song-server/src/main/java/bio/overture/song/server/model/entity/AnalysisSchema.java
+++ b/song-server/src/main/java/bio/overture/song/server/model/entity/AnalysisSchema.java
@@ -8,9 +8,11 @@ import bio.overture.song.server.model.analysis.Analysis;
 import bio.overture.song.server.model.enums.ModelAttributeNames;
 import bio.overture.song.server.model.enums.TableAttributeNames;
 import bio.overture.song.server.model.enums.TableNames;
+import bio.overture.song.server.utils.StringListConverter;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.databind.JsonNode;
 import java.time.LocalDateTime;
+import java.util.List;
 import java.util.Set;
 import javax.persistence.*;
 import javax.validation.constraints.NotNull;
@@ -52,8 +54,9 @@ public class AnalysisSchema {
   @Type(type = CUSTOM_JSON_TYPE_PKG_PATH)
   private JsonNode schema;
 
-  @Column(name = FILE_TYPES)
-  private String fileTypes;
+  @Column(name = FILE_TYPES, columnDefinition = "text[]")
+  @Convert(converter = StringListConverter.class)
+  private List<String> fileTypes;
 
   @JsonIgnore
   @Builder.Default
--- a/song-server/src/main/java/bio/overture/song/server/model/entity/AnalysisSchema.java
+++ b/song-server/src/main/java/bio/overture/song/server/model/entity/AnalysisSchema.java
@@ -1,9 +1,6 @@
 package bio.overture.song.server.model.entity;
 
-import static bio.overture.song.server.model.enums.TableAttributeNames.ID;
-import static bio.overture.song.server.model.enums.TableAttributeNames.NAME;
-import static bio.overture.song.server.model.enums.TableAttributeNames.SCHEMA;
-import static bio.overture.song.server.model.enums.TableAttributeNames.VERSION;
+import static bio.overture.song.server.model.enums.TableAttributeNames.*;
 import static bio.overture.song.server.repository.CustomJsonType.CUSTOM_JSON_TYPE_PKG_PATH;
 import static com.google.common.collect.Sets.newHashSet;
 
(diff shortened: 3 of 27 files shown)


pom.xml
```diff
@@ -56,11 +56,14 @@
...
-            <name>Spring Releases</name>
-            <url>https://repo.spring.io/libs-release</url>
-        </repository>
+       <repository>
+          <id>lib-releases</id>
+          <name>Spring Lib Releases</name>
+          <url>https://repo.spring.io/libs-release</url>
+          <snapshots>
+              <enabled>false</enabled>
+          </snapshots>
```
joneubank: 👍

song-core/src/main/java/bio/overture/song/core/model/Options.java
```diff
@@ -0,0 +1,12 @@
+package bio.overture.song.core.model;
+
+import java.util.List;
+import lombok.*;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class Options {
```
joneubank: We need a more precise name for this, there could be lots of Options objects. Could we call it `AnalysisTypeOptions`?

song-server/src/main/java/bio/overture/song/server/model/entity/AnalysisSchema.java
```diff
@@ -63,6 +52,9 @@ public class AnalysisSchema {
   @Type(type = CUSTOM_JSON_TYPE_PKG_PATH)
   private JsonNode schema;
 
+  @Column(name = FILE_TYPES)
+  private String fileTypes;
```
joneubank: Are we storing the value as a delimited string? whats the delimiter? Is this documented anywhere? Is it possible to store a `List<String>`?

song-server/src/main/resources/db/migration/V1__Base_version.sql
```diff
@@ -49,7 +49,7 @@ CREATE TYPE specimen_type as ENUM(
 DROP TYPE IF EXISTS sample_type CASCADE;
 CREATE TYPE sample_type as ENUM('DNA','FFPE DNA','Amplified DNA','RNA','Total RNA','FFPE RNA');
 DROP TYPE IF EXISTS file_type CASCADE;
-CREATE TYPE file_type as ENUM('FASTA','FAI','FASTQ','BAM','BAI','VCF','TBI','IDX','XML','TGZ');
+--CREATE TYPE file_type as ENUM('FASTA','FAI','FASTQ','BAM','BAI','VCF','TBI','IDX','XML','TGZ');
```
joneubank: We shouldn't be modifying old migrations. We need to add new migration that modify our database so that those with an existing DB can update to the latest.

song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
```diff
@@ -244,11 +255,82 @@ private void validateAnalysisTypeSchema(JsonNode analysisTypeSchema) {
...
+
+    List<AnalysisSchema> schemaList = analysisSchemaRepository.findByName(analysisTypeName);
+    Optional<AnalysisSchema> oldSchemaOptional = Optional.empty();
+
+    if (CollectionUtils.isNotEmpty(schemaList)) {
+      oldSchemaOptional =
+          schemaList.stream()
+              .sorted(Comparator.comparing(AnalysisSchema::getCreatedAt).reversed())
+              .findFirst();
+    }
```
joneubank: If you're trying to find the latest stored schema, we should be using the version number, not the created at date. These will likely be the same but are not the same meaning. We have a function already for [getLatestVertsionNumber](https://github.com/overture-stack/SONG/blob/094fbfa4046d2a2bb2ed897729e19b7226530ac3/song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java#L204) in this very file. You can then get the schema with `analysisSchemaRepository.findByNameAndVersion(name, resolvedVersion)`.

song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
```diff
@@ -244,11 +255,82 @@ private void validateAnalysisTypeSchema(JsonNode analysisTypeSchema) {
...
+    String fileTypes = "";
+
+    if (options != null && CollectionUtils.isNotEmpty(options.getFileTypes()))
+      fileTypes = String.join(",", options.getFileTypes());
```
joneubank: I would prefer if we use brackets to wrap all if statement bodies. This helps clarify what an if enables for future readers and prevents accidental bugs being introduced by future editors:

```suggestion
    if (options != null && CollectionUtils.isNotEmpty(options.getFileTypes())) {
      fileTypes = String.join(",", options.getFileTypes());
    }
```

song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
```diff
@@ -244,11 +255,82 @@ private void validateAnalysisTypeSchema(JsonNode analysisTypeSchema) {
...
+              .build();
+
+    } else {
+      analysisSchema =
+          AnalysisSchema.builder()
+              .name(analysisTypeName)
+              .schema(newAnalysisTypeSchema)
+              .fileTypes(fileTypes)
+              .build();
+    }
```
joneubank: Duplicate code in your AnalysisSchema builder. Would prefer that you complete building the schema separately and then only perform the analysisSchema declaration and instantiation once.

Something like:
```java
		val schemaJsonNode = oldSchemaOptional.isPresent() ? mergeSchema(oldAnalysisSchema.get().getSchema(), newAnalysisTypeSchema) : newAnalysisTypeSchema;
		AnalysisSchema analysisSchema = AnalysisSchema.builder()
			.name(analysisTypeName)
			.schema(schemaJsonNode)
			.fileTypes(fileTypes)
			.build();
```

song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
```diff
@@ -244,11 +255,82 @@ private void validateAnalysisTypeSchema(JsonNode analysisTypeSchema) {
     }
   }
 
+  private JsonNode mergeSchema(JsonNode oldSchema, JsonNode newSchema) {
```
joneubank: Why do we need this function? We weren't previously merging old schemas with new schemas, why is this needed now?

This function does not appear to have anything to do with AnalysisTypes other than that they are JsonNodes that need to be merged (if they do need to be merged...). Should this then be in a class for JSON utilties?

Lastly, is this code copied from somewhere else? This looks like a reusable method for merging JsonNodes, so if we did get this from somewhere else we should give it attribution to the original source. Please also update it to the coding format of the song codebase, for instance using lombok style variable declarations (`val`, `var`).

song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
```diff
@@ -244,11 +255,82 @@ private void validateAnalysisTypeSchema(JsonNode analysisTypeSchema) {
...
+    if (oldSchemaOptional.isPresent()) {
+
+      val oldAnalysisSchema = oldSchemaOptional.get();
+      val mergedSchemaJsonNode = mergeSchema(oldAnalysisSchema.getSchema(), newAnalysisTypeSchema);
+      analysisSchema =
+          AnalysisSchema.builder()
+              .name(analysisTypeName)
+              .schema(mergedSchemaJsonNode)
+              .fileTypes(fileTypes)
+              .build();
```
joneubank: Why do we need to do this merging logic that wasn't present before?

song-server/src/main/java/bio/overture/song/server/service/ValidationService.java
```diff
@@ -85,6 +90,27 @@ public Optional<String> validate(@NonNull JsonNode payload) {
...
+            String fileName = file.get("fileName").asText();
+            if (!fileTypes.contains(fileType)) {
+              throw new ValidationException(
+                      String.format("%s name is not supported, supported formats are %s"
+                              , fileName
+                              , String.join(", ", fileTypes)));
+            }
+          }
+        }
+      }
```
joneubank: We should pull this logic into its own function to `validateFileType(payload, analysisType)`, or at very least leave some comments to explain what this section of code is doing.

Also, lets do this after we have run `validateWithSchema()` since that will ensure the analysis has the correct properties at all.

joneubank (changes requested):

song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
```diff
@@ -246,9 +256,23 @@ private void validateAnalysisTypeSchema(JsonNode analysisTypeSchema) {
...
+    }
     val analysisSchema =
-        AnalysisSchema.builder().name(analysisTypeName).schema(analysisTypeSchema).build();
+        AnalysisSchema.builder()
+            .name(analysisTypeName)
+            .schema(analysisTypeSchema)
+            .fileTypes(fileTypes)
+            .build();
+
+    log.info("file types " + fileTypes);
```
joneubank: This looks like developer test logs, I think we can remove this or change it to debug level and add more details from the analysis schema that is being created.

song-server/src/main/resources/db/migration/V1_20__update_file_from_enum.sql
joneubank, line 1: 👍

song-server/src/test/java/bio/overture/song/server/controller/AnalysisTypeControllerTest.java
```diff
@@ -849,7 +850,7 @@ public static List<AnalysisType> generateData(
             i -> {
               val name = names.get(i % repeats);
               val schema = generateRandomRegistrationPayload(randomGenerator);
-              val out = analysisTypeService.register(name, schema);
+              val out = analysisTypeService.register(name,new AnalysisTypeOptions(), schema);
```
joneubank: Looks like code formatting hasn't been run on this line, no space between comma and next argument.

joneubank (changes requested):
This looks complete, just two small nitpicks.

Can you add a comment to this PR that describes the behaviour when we are creating a new version of an existing schema - if you don't specify file types does it keep the existing file type restrictions or remove the existing restrictions? We can create another ticket to update this behaviour if it is unintuitive, I just want to record what the behaviour is.

joneubank (approved):
Looking good!

Issue #875, opened 2024-10-24
Perform External Validation on Analysis


joneubank:
## Summary of request
Building off of https://github.com/overture-stack/SONG/issues/871 and https://github.com/overture-stack/SONG/issues/872

Now that we have external validation rules stored for an analysis schema, we need to perform that external validation when an analysis is submitted or updated.

## Details
- when creating or updating an analysis, we will check for external validation rules that are attached to the analysis type
- check for and apply external validations after the analysis passes the JSON schema validation and the file type validations (if present)
- for each external validation rule on this analysis type:
   - check the value of the property in the analysis at the given path
   - if there is no value defined at that path, then no validation is required, this passes
   - if there is a value defined at that path, then we will check it with the configured external source
   - build a URL request to the URL included in the external validation rule

### Request to the External Validation URL
Request definition:
- GET Request
- use the URL exactly as provided in analysis type config.
- add query parameters:
   - `studyId` with the value of the study the analysis belongs to
   - `value` with the value as taken from the analysis property being validated

This URL will then look something like: `${url}?studyId=${study ID from analysis}&value=${property value from analysis}`

Expected response
- if the value is valid according to the external source, the response will have HTTP status `200` with no body
- if value is invalid, the external source will return HTTP status `400`. The external server may optionally return a JSON with the following form:
   ```json
   {
      "error": "INVALID_VALUE",
      "message": "Any string message describing why the value is invalid"
   }
   ```
   The message from this response can be included in the response to the analysis submitter as an explanation for why the provided analysis is invalid.
- the external service may return other HTTP status error codes, for instance if it does not recognize the request or believes it to be malformed. SONG must be prepared for these errors, and when returned we will treat them as a failed validation. When this occurs, a message should be returned to the analysis submitter explaining that an external validation on their analysis was unable to complete and the analysis could not be submitted/updated.

## Additional context
This is meant to integration with [Lyric](https://github.com/overture-stack/lyric). Corresponding details for the development of this songValidation endpoint on that service will be linked from here once available.

Issue #872, opened 2024-10-21
Store External Validation Rules for Dynamic Schemas


joneubank:
Dependent on completion of: https://github.com/overture-stack/SONG/issues/871

## Summary of External Validation
We are adding a mechanism to validate that a property of an analysis, for example a donor ID, is registered in an external registry. This validation will be done during Analysis submission by making a web request to a configurable URL, and it will ensure that the value of a property is valid according to the external service.

External validation may be done to ensure that an ID is registered in another meta data service, or could be looking up a value in an external ontology, or any other validation logic that the external service chooses.

## Ticket Details
Dynamic Schemas will now contain external validation rules and these need to be stored in the DB for their analysis type.

Each analysis type could have no external validation rules, or multiple external valdation rules.
Each validation rule contains:
   - a string for the jsonPath to the property that will be validated
   - a string for the url that will be used for external validation

## Desired solution
_Please describe changes to data model here._

PR #877, opened 2024-10-31
Add External Validation Rules for Analysis Properties Based on Dynamic Schema

Azher2Ali:
This PR introduces external validation capabilities for analysis properties, allowing for dynamic schemas to specify validation rules that are stored in the database. Each rule specifies a `jsonPath` to identify the property for validation and a URL to an external service for validation logic. During analysis submission, the system will validate specific property values by querying the external service, confirming that the values (e.g., donor IDs) meet the criteria defined by the external registry or ontology.

- Adds support for configurable external validation rules per analysis type.

- Enables database storage of validation rules tied to specific analysis types.

- Implements web request calls to validate property values according to the rules during submission.

This update also introduces external validation checks for analyses during creation and update in SONG. These checks are applied after JSON schema validation and file type validation. For each configured external validation rule on an analysis type, the system will:

**Check Analysis Properties:** Determine if the property at the specified path has a value. If no value exists, the validation is considered passed.

**Validate Using External Source:** If a value is present, SONG constructs a GET request to the external validation URL. The URL includes query parameters:

- studyId: the study identifier the analysis is associated with.

- value: the value of the property being validated.

**Process the External Validation Response:**

- If the value is valid, the external service returns an HTTP 200 status with no response body.

- If invalid, the external service returns an HTTP 400 status and optionally includes a JSON error message explaining why the value is invalid. This message is provided to the submitter for clarity.

- Other HTTP errors indicate a validation failure, and a message is returned to the submitter stating that external validation was unsuccessful, preventing the analysis submission/update.

This update sets the groundwork for calling external validation services, but the actual call to the external service is pending final implementation .We added this feature with an exception that the actual call to the external service is skipped as the service is not up and running and if we start invoking it, this will fail all the validations  #872 #875

### DIFF
--- a/song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
+++ b/song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
@@ -289,6 +289,9 @@ public class AnalysisTypeService {
       }
     }
 
+    if(options!=null)
+      options.setFileTypes(fileTypes);
+
     val analysisSchema =
         AnalysisSchema.builder()
             .name(analysisTypeName)
@@ -342,17 +345,13 @@ public class AnalysisTypeService {
 
   private AnalysisType convertToAnalysisType(
       AnalysisSchema analysisSchema, boolean hideSchema, boolean unrenderedOnly) {
-    AnalysisTypeOptions options = new AnalysisTypeOptions();
-    if (analysisSchema.getOptions() != null) {
-      options.setFileTypes(analysisSchema.getOptions().getFileTypes());
-      options.setExternalValidation(analysisSchema.getOptions().getExternalValidation());
-    }
+
     return AnalysisType.builder()
         .name(analysisSchema.getName())
         .version(analysisSchema.getVersion())
         .createdAt(analysisSchema.getCreatedAt())
         .schema(resolveSchemaJsonView(analysisSchema.getSchema(), unrenderedOnly, hideSchema))
-        .options(options)
+        .options(analysisSchema.getOptions())
         .build();
   }
 
--- a/song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
+++ b/song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
@@ -269,9 +269,8 @@ public class AnalysisTypeService {
     }
 
     // checking if file types is empty
-    // if the version is new version of the schema , we are checking the previous version allowed
-    // file types
-    // if it is new then it is empty
+    // if the analysisSchemaVersion is new version of the schema and fileTypes is empty then,
+    // we are checking the previous version to map the fileTypes allowed to the latestVersion
     if (fileTypes.isEmpty()) {
       List<AnalysisSchema> analysisSchemaList =
           analysisSchemaRepository.findAllByName(analysisTypeName);
@@ -289,8 +288,7 @@ public class AnalysisTypeService {
       }
     }
 
-    if(options!=null)
-      options.setFileTypes(fileTypes);
+    if (options != null) options.setFileTypes(fileTypes);
 
     val analysisSchema =
         AnalysisSchema.builder()
--- a/song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
+++ b/song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
@@ -17,11 +17,7 @@
 
 package bio.overture.song.server.service;
 
-import static bio.overture.song.core.exceptions.ServerErrors.ANALYSIS_TYPE_NOT_FOUND;
-import static bio.overture.song.core.exceptions.ServerErrors.ILLEGAL_ANALYSIS_TYPE_NAME;
-import static bio.overture.song.core.exceptions.ServerErrors.MALFORMED_JSON_SCHEMA;
-import static bio.overture.song.core.exceptions.ServerErrors.MALFORMED_PARAMETER;
-import static bio.overture.song.core.exceptions.ServerErrors.SCHEMA_VIOLATION;
+import static bio.overture.song.core.exceptions.ServerErrors.*;
 import static bio.overture.song.core.exceptions.ServerException.buildServerException;
 import static bio.overture.song.core.exceptions.ServerException.checkServer;
 import static bio.overture.song.core.utils.CollectionUtils.isCollectionBlank;
@@ -29,10 +25,7 @@ import static bio.overture.song.core.utils.JsonUtils.readTree;
 import static bio.overture.song.core.utils.Separators.COMMA;
 import static bio.overture.song.server.controller.analysisType.AnalysisTypeController.REGISTRATION;
 import static bio.overture.song.server.repository.specification.AnalysisSchemaSpecification.buildListQuery;
-import static bio.overture.song.server.utils.JsonSchemas.PROPERTIES;
-import static bio.overture.song.server.utils.JsonSchemas.REQUIRED;
-import static bio.overture.song.server.utils.JsonSchemas.buildSchema;
-import static bio.overture.song.server.utils.JsonSchemas.validateWithSchema;
+import static bio.overture.song.server.utils.JsonSchemas.*;
 import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static java.util.Objects.isNull;
@@ -156,17 +149,12 @@ public class AnalysisTypeService {
         resolveSchemaJsonView(analysisSchema.getSchema(), unrenderedOnly, false);
 
     AnalysisTypeOptions options = analysisSchema.getOptions();
-
-    List<String> fileTypes =
-        (options.getFileTypes() != null && !options.getFileTypes().isEmpty())
-            ? options.getFileTypes()
-            : new ArrayList<>();
     return AnalysisType.builder()
         .name(analysisTypeId.getName())
         .version(analysisTypeId.getVersion())
         .createdAt(analysisSchema.getCreatedAt())
         .schema(resolvedSchemaJson)
-        .options(AnalysisTypeOptions.builder().fileTypes(fileTypes).build())
+        .options(options)
         .build();
   }
 
--- a/song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
+++ b/song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
@@ -155,9 +155,11 @@ public class AnalysisTypeService {
     val resolvedSchemaJson =
         resolveSchemaJsonView(analysisSchema.getSchema(), unrenderedOnly, false);
 
+    AnalysisTypeOptions options = analysisSchema.getOptions();
+
     List<String> fileTypes =
-        (analysisSchema.getFileTypes() != null && !analysisSchema.getFileTypes().isEmpty())
-            ? analysisSchema.getFileTypes()
+        (options.getFileTypes() != null && !options.getFileTypes().isEmpty())
+            ? options.getFileTypes()
             : new ArrayList<>();
     return AnalysisType.builder()
         .name(analysisTypeId.getName())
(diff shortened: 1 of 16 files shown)


song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
```diff
@@ -265,11 +267,33 @@ private AnalysisType commitAnalysisType(
...
+
+      if (!analysisSchemaList.isEmpty()) {
+        Optional<AnalysisSchema> latestSchemaOptional =
+            analysisSchemaList.stream()
+                .filter(schema -> schema.getVersion() != null)
+                .max(Comparator.comparingInt(AnalysisSchema::getVersion));
+
+        if (latestSchemaOptional.isPresent()) {
+          AnalysisTypeOptions optionsFromDb = latestSchemaOptional.get().getOptions();
+          fileTypes = optionsFromDb.getFileTypes();
```
leoraba, line 274: Looks like this check is retrieving the previous `fileTypes` and not used except for logging.

song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
```diff
@@ -319,7 +343,10 @@ public static AnalysisTypeId resolveAnalysisTypeId(@NonNull AnalysisType analysi
   private AnalysisType convertToAnalysisType(
       AnalysisSchema analysisSchema, boolean hideSchema, boolean unrenderedOnly) {
     AnalysisTypeOptions options = new AnalysisTypeOptions();
-    options.setFileTypes(analysisSchema.getFileTypes());
+    if (analysisSchema.getOptions() != null) {
+      options.setFileTypes(analysisSchema.getOptions().getFileTypes());
+      options.setExternalValidation(analysisSchema.getOptions().getExternalValidation());
+    }
```
leoraba: This setters using getters seems a bit redundant. Guess the variable `options` is not required as we can use `analysisSchema.getOptions()` directly bellow in the code when using the Builder.

song-server/src/main/java/bio/overture/song/server/service/ValidationService.java
```diff
@@ -62,23 +68,26 @@ public class ValidationService {
...
     this.analysisTypeService = analysisTypeService;
     this.uploadRepository = uploadRepository;
     this.enforceLatest = enforceLatest;
     this.analysisTypeIdSchema = analysisTypeIdSchemaSupplier.get();
+    this.restTemplate = restTemplate;
   }
 
   @SneakyThrows
-  public Optional<String> validate(@NonNull JsonNode payload) {
+  public Optional<String> validate(@NonNull JsonNode payload, @NonNull String studyId) {
```
leoraba: `studyId` variable is not used in this method.

song-server/src/test/java/bio/overture/song/server/service/MockedSubmitTest.java
```diff
@@ -101,7 +101,7 @@ public void submit_StudyDne_StudyIdDoesNotExist() {
     // Verify
     assertSongError(
         () -> submitService.submit("anyStudy", "anyAnalysisId", false), STUDY_ID_DOES_NOT_EXIST);
-    verify(validationService, never()).validate(isA(JsonNode.class));
+    verify(validationService, never()).validate(isA(JsonNode.class),"anyStudy");
```
leoraba: what's the reason we're passing an study to the method `validate`?

song-server/src/main/java/bio/overture/song/server/service/ValidationService.java
```diff
@@ -109,6 +118,43 @@ public Optional<String> validate(@NonNull JsonNode payload) {
...
+      String value = JsonPath.read(payload, "$." + jsonPath);
+
+      if(Objects.isNull(value) || value.isEmpty())
+        return;
+
+      // call the external url using restTemplate
+    }
+  }
+
+  public boolean invokeExternalUrl(String studyId, String url, String value) {
```
leoraba: Don't see where are we calling this method.

song-server/src/main/java/bio/overture/song/server/service/ValidationService.java
```diff
@@ -62,23 +68,26 @@ public class ValidationService {
   private final UploadRepository uploadRepository;
   private final boolean enforceLatest;
   private final Schema analysisTypeIdSchema;
+  private final RestTemplate restTemplate;
```
leoraba: is there a way to use rather a `RetryTemplate`? just in case we require retry mechanism

leoraba (commented):
So far so good! Just left some comments to consider 📝

song-server/src/main/java/bio/overture/song/server/service/ValidationService.java
Azher2Ali: We are not calling this method as of now, because the external API is not up and running due to which all the validations will fail.

song-server/src/test/java/bio/overture/song/server/service/MockedSubmitTest.java
Azher2Ali: As we need to populate `studyId` value for external validation, hence we have added this as a parameter. If required "`anyStudy`" to "`anyStudyId`"

song-server/src/main/java/bio/overture/song/server/service/ValidationService.java
Azher2Ali: It's not used now, but this will be used when we invoke external URL method

song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
Azher2Ali: You are right, will change this part of code.

song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
Azher2Ali: We will be setting the `fileTypes` fetched from previous versions to `options `object

song-server/src/main/java/bio/overture/song/server/service/ValidationService.java
```diff
@@ -109,6 +118,43 @@ public Optional<String> validate(@NonNull JsonNode payload) {
...
+
+      // create URL
+      URIBuilder uriBuilder = new URIBuilder(url);
+      uriBuilder.addParameter("studyId", studyId);
+      uriBuilder.addParameter("value",value);
+      // Make the HTTP GET request
+      ResponseEntity<Void> response = restTemplate.getForEntity(uriBuilder.build().toURL().toString(), Void.class);
+
+      // Return true if the response status is 200 OK
+      return response.getStatusCode().is2xxSuccessful();
```
joneubank: 👍

song-server/src/main/java/bio/overture/song/server/service/ValidationService.java
Azher2Ali: In order to apply Retry mechanism on the `RestTemplate` we should consider Timeout on UI as retry will take considerable amount of time. As a result, if the `RestTemplate` fails. means that the whole operations failed.

song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
```diff
@@ -265,11 +267,36 @@ private AnalysisType commitAnalysisType(
     if (options != null && CollectionUtils.isNotEmpty(options.getFileTypes())) {
       fileTypes = options.getFileTypes();
     }
+
+    // checking if file types is empty
+    // if the version is new version of the schema , we are checking the previous version allowed
+    // file types
+    // if it is new then it is empty
```
joneubank: This comment is confusing, I don't understand what it is trying to say. Can you review and rephrase it to describe the purpose of the following section?

song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
```diff
@@ -265,11 +267,36 @@ private AnalysisType commitAnalysisType(
...
+            analysisSchemaList.stream()
+                .filter(schema -> schema.getVersion() != null)
+                .max(Comparator.comparingInt(AnalysisSchema::getVersion));
+
+        if (latestSchemaOptional.isPresent()) {
+          AnalysisTypeOptions optionsFromDb = latestSchemaOptional.get().getOptions();
+          fileTypes = optionsFromDb.getFileTypes();
+        }
+      }
+    }
```
joneubank, line 277: Why does this PR update how the `fileType` options are being set? Wasn't this functionality already in place?

Do we perform the same logic for `externalValidation` rules? (When an analysis is being updated, if no external validation rules are provided we copy the externalValidaiton rules from the latest schema version)

song-server/src/main/java/bio/overture/song/server/service/ValidationService.java
```diff
@@ -100,6 +119,7 @@ public Optional<String> validate(@NonNull JsonNode payload) {
         validateFileType(fileTypes, payload);
       }
 
+      // log.info("SCHEMA :- " + analysisType.getSchema());
```
joneubank: Remove commented code.

song-server/src/main/java/bio/overture/song/server/service/ValidationService.java
```diff
@@ -109,6 +129,65 @@ public Optional<String> validate(@NonNull JsonNode payload) {
...
+      validated = invokeExternalUrl(studyId, externalUrl, value);
+    }
+    return validated;
+  }
+
+  public boolean invokeExternalUrl(String studyId, String url, String value) {
+    try {
+
+      Pattern pattern = Pattern.compile(LYRIC_URL_REGEX);
+      Matcher matcher = pattern.matcher(url);
```
joneubank: Variable names should describe their content, not just `pattern` and `matcher` but `lyricUrlPattern` and `lyricUrlMatcher` or similar.

song-server/src/main/java/bio/overture/song/server/service/ValidationService.java
```diff
@@ -109,6 +129,65 @@ public Optional<String> validate(@NonNull JsonNode payload) {
...
+      // call the external url using restTemplate
+      validated = invokeExternalUrl(studyId, externalUrl, value);
+    }
+    return validated;
+  }
+
+  public boolean invokeExternalUrl(String studyId, String url, String value) {
+    try {
+
+      Pattern pattern = Pattern.compile(LYRIC_URL_REGEX);
```
joneubank: What about external validation URLs when they are validating against a service that isn't Lyric? We need to define a schema for writing a URL that isn't specific to Lyric. This is about defining string template rules with variables we will replace with the studyId and entity name. 

Perhaps `${studyId}`, `${entityName}`, and `${entityId}`. So an external URL value could be `https://mydomain.example.org/whatever/path/to/validate/${studyId}/${entityName}/${entityId}`.

song-server/src/main/resources/db/migration/V1_21__remove_file_types_and_add_options.sql
```diff
@@ -0,0 +1,8 @@
+
+-- Remove the file_types column
+ALTER TABLE public.analysis_schema
+DROP COLUMN file_types;
```
joneubank: Have we published this version with file types yet? We would need to migrate all the file_types that have already been set into the new analysis schema format.

Did you consider storign analysis_schema in a separate column instead of storing options as a JSON?

joneubank (changes requested):

song-server/src/main/java/bio/overture/song/server/service/ValidationService.java
Azher2Ali: The code will take any URL apart from lyric, but for the path and path params should be well defined in order to have a common logic

song-server/src/main/resources/db/migration/V1_21__remove_file_types_and_add_options.sql
joneubank: I think that in the long term it would be best to store all options as their own dedicated columns in postgres. For the scope of this PR, we'll just leave it as a column for options since this is working today.

Issue #863, opened 2024-10-02
Epic - Separate Donor/Specimen/Sample model from Analysis model


joneubank:
## Summary
Song should be usable for files which do not use the Donor/Specimen/Sample model. The current Analysis data model makes these fields a requirement for every submitted analysis, forcing data that does not use these fields to fill in these fields to satisfy the software. Additionally, when multiple analyses are submitted for a single donor, the donor information needs to be repeated for each of these analyses. This causes a duplication of data, and can even suffer from input errors from one analysis to another. There is an additional issue with Song data model for Donors/Specimen/Samples being limited and not customizable to the data that different systems wish to collect.

Overture is developing towards having separate services for tracking structured data, see [Lectern](https://github.com/overture-stack/lectern) and [Lyric](https://github.com/overture-stack/lyric). These services provide the ability to track any data model for Donors, registration of their Specimen and Samples, and any other clinical or phenotypical data that is relevant to a study. This frees up Song to focus on being the service to track Analysis meta-data.

In order to connect the Analysis data with related structured data, a system can include a field in their Dynamic Schema which will provide an ID to link this analysis to the data tracked in lyric. This mapping becomes fully customizable through the Dynamic Schema definition. With this change, we will need to provide a mechanism for Song to check with an external service to validate that the provided value in one of these fields is registered with an external ID/data service.

### Song as ID Service
One feature of Song that is lost by this change will be the use of Song to generate system wide unique IDs for Donors, Specimen, and Samples. Song has previously had the option to work as an ID server, generating unique IDs for these entities the first time an analysis that referenced them was submitted. As these entities will no longer be a standard part of the Analysis model, there is no mechanism for identifying Donors/Speciment/Samples and so they will no longer be tracked by Song's database, and no Donor/Specimen/Sample IDs will be generated by Song.

Most cases where Song has been used do not require this feature. A separate ID service has been used to create system wide IDs. This is either because of a federated data management model that has multiple Song instances, or because clinical data records are being registered and tracked separate from the file data. This will standardize this process for all future instances of Song.

### Updating the Data Model
To accomplish this change we need to remove these entities from several parts of the code base, and update some functionality of the Song Server.

- [x] https://github.com/overture-stack/SONG/issues/864 
- [x] https://github.com/overture-stack/SONG/issues/865
- [x] Update AnalysisService interface and implementations to not include any references to Donor, Specimen, or Sample specific code.
   - Includes updating PayloadConverter
[...]

PR #878, opened 2025-01-22
Refactor Base Schema to not require Donor/Specimen/Sample

joneubank:
## Issue Reference
- https://github.com/overture-stack/SONG/issues/863

This change removes all references to donors, specimens, and samples from the base schema. Analyses now have very little required information, primarily the Study and list of Files. Since Analyses no longer require donor or sample information, the relation between donors, specimens, and samples is no longer built and validated during analysis submission. See the [linked issue](https://github.com/overture-stack/SONG/issues/863) for details on the motivation for this change.

## Changes
- Remove Donor/Specimen/Sample properties from base analysis schema
- Remove search by Donor/Specimen/Sample IDs from the Analysis ID Search API since these aren't default properties of analyses
- Add External Validation options on custom schemas and validation , to replace the ID Server checks that are not possible now that Donor and Samples are removed from the base schema
   - https://github.com/overture-stack/SONG/pull/877
   - https://github.com/overture-stack/SONG/pull/879

## Bug Fixes
The following fixes are included in this work but not captured in the above tickets
- Paginated search was broken due to the fileType enum removal. The stored procedure for `get_analysis` is updated to no longer join with the donor/specimen/sample tables and fixes this type issue
- Analysis schema options properties, including `fileTypes` are set to empty lists by default, and can be cleared by providing an empty list.


docs/custom-schemas.md
```diff
@@ -0,0 +1,97 @@
...
+     "properties":{
+        "experiment":{}
+     }
+  }
+}
+```
+
+## Options
+
+All properties in options are optional. If no value is provided, a default configuration will be used for the analysis. If this is an update to an existing analysis type, you can omit any option and the value will be maintained from the previous version.
```
justincorrigible:
```suggestion
All properties in "options" are elective. If no value is provided, a default configuration (shown bellow) will be used for the analysis. If this is an update to an existing analysis type, you can omit any option and its value will be maintained from the previous version (if one is available).
```
is the following object a representation of the default options? added note here in case that is so, to make it explicit. otherwise, we could include a link to the file/code that contains said default options.

docs/custom-schemas.md
```diff
@@ -0,0 +1,97 @@
...
+{
+	"options": {
+		"fileTypes": ["bam","cram"]
+	}
+}
+```
+
+### External Validation
+
+External validations configure Song to check a value in the analysis against an external service by sending an HTTP GET request to a configurable URL. The URL needs to return 2XX status message to indicate that if the provided value "is valid", typically meaning that the value for this property is known by that service.
```
justincorrigible:
```suggestion
External validations configure Song to check a value in the analysis against an external service, by sending an HTTP GET request to a configurable URL. The service should respond with a 2XX status message to indicate the value "is valid".
```

docs/custom-schemas.md
```diff
@@ -0,0 +1,97 @@
...
+      "fileMd5sum": "595f44fec1e92a71d3e9e77456ba80d1"
+    }
+  ],
+  "experiment": {
+    "donorId": "id01"
+  }
+}
+```
+
+Song would attempt to validate the donorId by sending a URL to `http://example.com/ABC123/donor/id01`.
```
justincorrigible:
```suggestion
Song would attempt to validate the donorId by sending a validation request to `http://example.com/ABC123/donor/id01`.
```

docs/custom-schemas.md
```diff
@@ -0,0 +1,97 @@
...
+
+It is allowable to use either the `{study}` or `{value}` multiple times in the URL, each instance will be replaced by the corresponding value.
```
justincorrigible:
```suggestion
The URL parsing allows using either the `{study}` or `{value}` placeholders multiple times (e.g. `http://example.com/{study}-{value}/{value}`), each instance will be interpolated accordingly.
```

docs/custom-schemas.md
```diff
@@ -0,0 +1,97 @@
...
+
+> [!Warning]
+> The URL may cause errors in song if it contains any tokens matching the `{word}` format other than `{study}` and `{value}`
```
justincorrigible: Changing this here for consistency with the rest of your additions... and now I realize I've been noticing we spell SONG/Song/song in different places (beyond this repo).
Is it worth making a pass in a different PR to unify those here? if so, which spelling should we use? in Overture docs, we've been using all caps. TBD

```suggestion
> The URL may cause errors in Song if it contains any tokens matching the `{word}` format other than `{study}` and `{value}`
```

song-client/src/main/java/bio/overture/song/client/command/SearchCommand.java
```diff
@@ -108,18 +89,15 @@ private boolean isAnalysisSearchMode() {
   }
 
   private boolean isIdSearchMode() {
-    return nonNull(fileId) || nonNull(sampleId) || nonNull(specimenId) || nonNull(donorId);
```
justincorrigible, line 111: 🎉

song-server/src/main/java/bio/overture/song/server/model/entity/AnalysisSchema.java
```diff
@@ -4,24 +4,18 @@
...
 import java.util.Set;
 import javax.persistence.*;
 import javax.validation.constraints.NotNull;
-import lombok.AllArgsConstructor;
-import lombok.Builder;
-import lombok.Data;
-import lombok.EqualsAndHashCode;
-import lombok.NoArgsConstructor;
-import lombok.ToString;
+import lombok.*;
```
justincorrigible, line 18: How's Maven's tree shaking these days? I recall this `.* pattern could add some pounds to the code back when

song-server/src/main/resources/schemas/analysis/analysisBase.json
justincorrigible, line 1: This makes me soooooo happy! 👏

song-server/src/test/java/bio/overture/song/server/utils/generator/PayloadGenerator.java
```diff
@@ -47,15 +47,7 @@ public class PayloadGenerator {
...
-    payload
-        .getSamples()
-        .forEach(
-            x -> {
-              x.setSubmitterSampleId(randomGenerator.generateRandomUUID().toString());
-              x.getSpecimen()
-                  .setSubmitterSpecimenId(randomGenerator.generateRandomUUID().toString());
-              x.getDonor().setSubmitterDonorId(randomGenerator.generateRandomUUID().toString());
-            });
+    ;
```
justincorrigible: typo?
```suggestion
```

justincorrigible (commented):
This is just awesome! long time coming too. 
thanks for taking the time to get this done!

Left a couple comments on the docs side, the rest of the changes LGTM 👍 (except for that random semicolon 😆 maybe I'm missing something?)

docs/custom-schemas.md
justincorrigible: also, if I omit an option because I'm trying to clear it for whatever reason, but it was already set (and therefor the existing value will be maintained), should we explain here how to remove it?

docs/custom-schemas.md
joneubank: Including all these changes in my next commit, some of them conflict with a change I made so I'll commit the suggestions but not through the GH UI.

song-server/src/main/java/bio/overture/song/server/model/entity/AnalysisSchema.java
joneubank: I'm not sure on this but I know the common practice for code formatters is that after a minimum number of imports from a package the formatter will switch to the `.*` form. So I would assume there is not a meaningful package size change associated with this. Important to know that this affects docker image size, less impactful than a web page bundle size increase.

joneubank:
I changed the optional property name from `externalValidation` to `externalValidations` to be consistent with naming of array properties. It matches `fileTypes` now (both plural).

docs/custom-schemas.md
```diff
@@ -0,0 +1,110 @@
...
+     "properties":{
+        "experiment":{}
+     }
+  }
+}
+```
+
+## Options
+
+The `options` property is not required, if provided extra validations for this analysis can be specified. Similarly, each property in `options` is also optional. If no value is provided for an option, a default configuration will be used for the analysis. If this is an update to an existing analysis type, you can omit any option and its value will be maintained from the previous version.
```
justincorrigible:
```suggestion
The `options` property is not required. If provided, extra validations for this analysis can be specified. Similarly, each property in `options` is also optional. If no value is provided for an option, a default configuration will be used for the analysis. If this is an update to an existing analysis type, you can omit any option and its value will be maintained from the previous version.
```

docs/custom-schemas.md
```diff
@@ -0,0 +1,110 @@
...
+			"url": "http://localhost:8099/",
+			"jsonPath": "experiment.someId"
+		}
+	]
+},
+```
+
+If you want to remove the previous value of an option so that this validation is no longer required, for instance removing the restriction on file types so that any file type could be provided, you should provide an empty list for that option.
+
+In the example below, both `fileTypes` and `externalValidations` properties are set to empty arrays, which means that these validations will not be applied to submitted analysis:
```
justincorrigible, line 39: love it. very helpful description 👍

justincorrigible (approved):

PR #879, opened 2025-01-27
Schema registrations sets options values correctly

joneubank:
- when missing, uses value from previous schema or sets as empty (default) value
- when set as empty array, replaces current value with empty array (removes optional validations)
- sets options in get registration by version query

### DIFF
--- a/song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
+++ b/song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
@@ -33,10 +33,7 @@ import static java.util.regex.Pattern.compile;
 import static org.apache.commons.lang.StringUtils.isBlank;
 
 import bio.overture.song.core.exceptions.ServerErrors;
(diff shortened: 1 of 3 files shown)


song-server/src/main/java/bio/overture/song/server/service/AnalysisTypeService.java
```diff
@@ -244,48 +243,57 @@ private void validateAnalysisTypeSchema(JsonNode analysisTypeSchema) {
...
+            ? options.getExternalValidation()
+            : previousOptions.getExternalValidation() != null
+                ? previousOptions.getExternalValidation()
+                : new ArrayList<>();
+
+    val newAnalysisOptions =
+        AnalysisTypeOptions.builder()
+            .fileTypes(fileTypes)
+            .externalValidation(externalValidations)
+            .build();
```
joneubank, line 288: This is the bulk of the change. We set values for both options, fileTypes and externalValidations, that will:
- check for value in input
- if no input value, check for value from previous version
- if no previous version, apply empty value (default)

PR #907, opened 2026-03-20
Fetch samples structure using batch JOIN query

leoraba:
# Description
This PR makes some changes to the Analysis Samples migration script.

## Changes:
- It replaces 3 individual queries (getSamplesById, getSpecimenById, getDonorById) with 1 single batch JOIN query.
- Added a Dockerfile to build a Docker image

### DIFF
--- a/data-migration/Dockerfile
+++ b/data-migration/Dockerfile
@@ -1,0 +1,86 @@
+# Global variables
+ARG COMMIT=""
+ARG APP_USER=data-migration
+ARG WORKDIR=/usr/src/app
+
+######################
+# Configure base image
+######################
+FROM node:24-alpine AS base
+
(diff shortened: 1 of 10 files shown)


justincorrigible:
PR converted to Draft, as it's pending changes in the jenkins pipeline so this new image won't be built unnecessarily (i.e. when other things in the repo change). current pipeline status just forks the job based on branch, which is not sustainable long term

leoraba:
The pipeline has been updated to by default only build Song.
 
If a new build for the data migration is required, it can be executed manually using Jenkins interface by selecting 
- Build Application: `Migration Script`
- Publish Image: `checked`

justincorrigible (approved):

Issue #880, opened 2025-01-28
Feature Request - External Validation should be able to test data in arrays


joneubank:
## Summary of request
Data from an array value and from value in an object nested in array should be able to be read and tested against an external service. In this scenario, we should be able to either pick the value from a specific postiion in an array, or test every value in the array against the external source

## Details
TODO

## Desired solution
TODO

## Additional context
TODO

joneubank:
Additional detail: The current implementation only allows external validation of properties of the analysis that are Strings. If the JSON Path given in the external validation points to a value that is not a string, then the external validation will fail (and the logs will indicate no value read for that property).

Buwujiu:
Merged to song, but not in file manager yet?

PR #909, opened 2026-07-24
Allow external validations to reference elements of an array

joneubank:
## Summary

External validations can reference elements in an array, or validate every element in an array.

Non-string and non-array values retrieved from analysis data will raise a `ValidationException`, causing the analysis to fail validation.

Added test dependency `junit-vintage-engine` to ensure all JUnit 4 defined tests (including those added to verify external validation functionality) are run with the mvn test command.

## Issues
- #880


song-server/src/main/java/bio/overture/song/server/service/ValidationService.java
```diff
@@ -127,14 +128,27 @@ public Optional<String> validate(@NonNull JsonNode payload, @NonNull String stud
     return Optional.ofNullable(errors);
   }
 
-  private Optional<String> getValueAtJsonPath(@NonNull JsonNode payload, @NonNull String jsonPath) {
+  List<String> getValuesAtJsonPath(@NonNull JsonNode payload, @NonNull String jsonPath) {
```
justincorrigible: `getValuesAtJsonPath` drops the `private` to package-private... was that to make it directly testable?
if so, worth adding that test in this PR. otherwise, might as well keep it private since nothing else in the package calls it

song-server/src/main/java/bio/overture/song/server/service/ValidationService.java
```diff
@@ -151,26 +165,25 @@ private void externalValidations(
 
     for (ExternalValidation externalValidation : externalValidations) {
 
-      val value = getValueAtJsonPath(payload, externalValidation.getJsonPath());
-      if (value.isPresent()) {
-        // Only validate vs external source if the value is present in the analysis payload
+      val values = getValuesAtJsonPath(payload, externalValidation.getJsonPath());
+      for (String value : values) {
```
justincorrigible, line 183: for a wildcard path like e.g. `donors[*].donorId`, this now makes one blocking `restTemplate.getForEntity` call per array element, sequentially, in the request thread, fail-fast on the first error, which is fine for small arrays, but could add bad latency for a study with a large donor list

for clarity: not asking to fix it here, but might be worth a follow-up ticket, or at least a line in the docs flagging the latency cost for larger arrays

song-server/src/main/java/bio/overture/song/server/service/ValidationService.java
```diff
@@ -127,14 +128,27 @@ public Optional<String> validate(@NonNull JsonNode payload, @NonNull String stud
...
+        for (val element : (List<?>) result) {
+          if (element instanceof String) {
+            values.add((String) element);
+          }
+        }
+        return values;
+      } else if (result instanceof String) {
+        return List.of((String) result);
+      } else {
+        return List.of();
```
justincorrigible, line 152: the old code's `String value = JsonPath.read(...)` (old line 132) would throw a `ClassCastException` on a non-string result, which got logged at debug in the catch block; this `else` branch here now takes the same case (non-string result) but returns silently, no log line unlike the one in line 151

if someone points jsonPath at the wrong type (e.g. an array of objects instead of strings), validation will just quietly no-op with no visibility.... wanna add a debug log here too? that way a misconfigured jsonPath is at least visible to someone looking at logs 🤞

justincorrigible (commented):
While it's a relatively tiny PR, it changes the branching logic quite a bit (list vs. scalar vs. non-string vs. exception, plus the new fail-fast loop over multiple values), but there's no tests added which would ideally check each conditional path for its own failure modes... and it looks like `ValidationService` has no test file at all yet (not entirely sure I may be misreading it, because java)

it would probably be easy to toss an agent at it so we can add a small/minimal ValidationServiceTest coverage: explicitly at least, a definite path returning one value, wildcard path returning multiple values, a wildcard result with a non-string element mixed in, an empty array, and one bad value aborting the rest of the loop?

that would also lock in the "empty array = no calls" and "one call per array element" behavior the docs now promise

song-server/src/main/java/bio/overture/song/server/service/ValidationService.java
joneubank: Good point here. 

Each incoming request in the Spring Boot app is handled on its own thread so I'm not worried about blocking the server like in a Node app, but this is a valid concern for response times on the create analysis request. I don't know if there is a fix here... in order to allow an admin to let every element of an array be tested, they will have to deal with the delay for testing every element in the array. I think I will add a note about that in the docs.

song-server/src/main/java/bio/overture/song/server/service/ValidationService.java
joneubank: ~~Same error thrown and now caught by ` } catch (Exception exception) {` on line 147. Same message is logged as before. Empty list is now returned.~~

`JsonPath.read` no longer immediately cast to string, the case you described is handled by the default `else` block on 144. I will add debug log.

song-server/src/test/java/bio/overture/song/server/validation/GetValuesAtJsonPathTest.java
```diff
@@ -0,0 +1,156 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package bio.overture.song.server.service;
```
leoraba:
```suggestion
package bio.overture.song.server.validation;
```

song-server/src/main/java/bio/overture/song/server/service/ValidationService.java
leoraba: in order to be accesible from the Test it needs to be `public`

song-server/src/test/java/bio/overture/song/server/validation/GetValuesAtJsonPathTest.java
```diff
@@ -0,0 +1,156 @@
...
+import org.everit.json.schema.ValidationException;
+import org.junit.Before;
+import org.junit.Test;
+import org.springframework.web.client.RestTemplate;
+
+public class GetValuesAtJsonPathTest {
+
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+
+  private ValidationService service;
```
leoraba, line 41: due to the suggested change pckage name it now needs to be imported from the right package

docs/02-Reference/04-data-model-management.md
```diff
@@ -169,6 +169,34 @@ The URL parsing allows using either the `{study}` or `{value}` placeholder multi
...
+```json
+{
+	"url": "http://example.com/{study}/donor/{value}",
+	"jsonPath": "donors[*].donorId"
+}
+```
+
+This extracts `donorId` from each element of the `donors` array and performs one validation call per value. If any value fails validation, the entire analysis submission is rejected. If the array is empty, no validation calls are made and the submission proceeds.
+
+Only string values are validated. If the property at the resolved path is not a string, it is silently skipped.
```
justincorrigible, line 198: this line still says "only string values are validated... it is silently skipped," but that's not what the code does anymore. maybe better now something like e.g. "the resolved value(s) must be strings; a non-string value fails validation for the whole submission."?

song-server/src/main/java/bio/overture/song/server/service/ValidationService.java
justincorrigible: just circling back on this one, you mentioned you'd add a note in the docs about the latency cost of validating every element in a large array, don't see that note in the current diff yet. 

still worth adding before merge, or happy to leave it as a fast follow if you'd rather not block on it. your call 👍

